PR #5, opened 2021-02-22
Add environment variable to not force asynchronous publish mode

ivanpauno:
Currently we're forcing asynchronous publish mode in all publishers for all rmw implementations that have the option available.
This might not be suitable for all use cases.

This PR adds an environment variable to avoid that behavior, so users can specify the publish mode in a qos profile file à la DDS way (if desired using topic filters, etc).

`rmw_connext_cpp` used a similar env variable (see [here](https://github.com/ros2/rmw_connext/blob/fd3e6a71fee0f7eb26f84505d16169ce489d6235/rmw_connext_shared_cpp/src/init.cpp#L117)).
`rmw_fastrtps_cpp` provides a similar env variable (see [here](https://github.com/ros2/rmw_fastrtps/blob/fe2fcb9c627a21433549314c209892fb5646b9b1/README.md#change-publication-mode)).

### DIFF
--- a/rmw_connextdds_common/src/common/rmw_context.cpp
+++ b/rmw_connextdds_common/src/common/rmw_context.cpp
@@ -184,6 +184,25 @@ rmw_context_impl_t::initialize_node(
 
   this->qos_library = qos_library;
 
+  /* Lookup name of custom QoS library */
+  const char * do_not_override_publish_mode_env = nullptr;
+  lookup_rc = rcutils_get_env(
+    RMW_CONNEXT_ENV_DO_NOT_OVERRIDE_PUBLISH_MODE, &do_not_override_publish_mode_env);
+
+  if (nullptr != lookup_rc || nullptr == qos_library) {
+    RMW_CONNEXT_LOG_ERROR_A_SET(
+      "failed to lookup from environment: "
+      "var=%s, "
+      "rc=%s ",
+      RMW_CONNEXT_ENV_DO_NOT_OVERRIDE_PUBLISH_MODE,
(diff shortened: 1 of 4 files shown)


asorbini (approved):

asorbini:
Thank you for this contribution @ivanpauno!

rmw_connextdds_common/src/common/rmw_context.cpp
```diff
@@ -184,6 +184,25 @@ rmw_context_impl_t::initialize_node(
 
   this->qos_library = qos_library;
 
+  /* Lookup name of custom QoS library */
+  const char * do_not_override_publish_mode_env = nullptr;
+  lookup_rc = rcutils_get_env(
+    RMW_CONNEXT_ENV_DO_NOT_OVERRIDE_PUBLISH_MODE, &do_not_override_publish_mode_env);
+
+  if (nullptr != lookup_rc || nullptr == qos_library) {
```
asorbini, line 192: I noticed after merging that this check should be for `do_not_override_publish_mode_env` and not `qos_library`. I can fix this in another commit, since I'm working on reorganizing the participant factory's initialization/finalization.

ivanpauno: Ups, wrong copy pasting

ivanpauno: I opened https://github.com/rticommunity/rmw_connextdds/pull/6 with a direct fix, but feel free to close that PR and commit the fix elsewhere is that easier.

PR #12, opened 2021-03-23
Add support for content-filtered topics

iuhilnehc-ynos:
related to ros2/design#282

Signed-off-by: Chen Lihui <lihui.chen@sony.com>

### DIFF
--- a/rmw_connextdds_common/include/rmw_connextdds/dds_api.hpp
+++ b/rmw_connextdds_common/include/rmw_connextdds/dds_api.hpp
@@ -208,6 +208,18 @@ rmw_connextdds_enable_security(
   rmw_context_impl_t * const ctx,
   DDS_DomainParticipantQos * const qos);
 
+rmw_ret_t
+rmw_connextdds_set_cft_filter_expression(
+  DDS_TopicDescription * const topic_desc,
+  const char * filter_expression,
+  const rcutils_string_array_t * expression_parameters);
+
+rmw_ret_t
+rmw_connextdds_get_cft_filter_expression(
+  DDS_TopicDescription * const topic_desc,
+  char ** const expr_out,
+  rcutils_string_array_t * cft_params_out);
+
 // Define some macro aliases for security-related properties
 #ifndef DDS_SECURITY_PROPERTY_PREFIX
 #define DDS_SECURITY_PROPERTY_PREFIX \
--- a/rmw_connextdds_common/include/rmw_connextdds/dds_api.hpp
+++ b/rmw_connextdds_common/include/rmw_connextdds/dds_api.hpp
@@ -72,6 +72,7 @@ rmw_connextdds_create_contentfilteredtopic(
   DDS_Topic * const base_topic,
   const char * const cft_name,
   const char * const cft_filter,
+  const rcutils_string_array_t * cft_expression_parameters,
   DDS_TopicDescription ** const cft_out);
 
 rmw_ret_t
--- a/rmw_connextdds_common/include/rmw_connextdds/dds_api.hpp
+++ b/rmw_connextdds_common/include/rmw_connextdds/dds_api.hpp
@@ -212,7 +212,7 @@ rmw_ret_t
 rmw_connextdds_set_cft_filter_expression(
   DDS_TopicDescription * const topic_desc,
   const char * const filter_expression,
-  const rcutils_string_array_t * expression_parameters);
+  const rcutils_string_array_t * const expression_parameters);
 
 rmw_ret_t
 rmw_connextdds_get_cft_filter_expression(
--- a/rmw_connextdds_common/include/rmw_connextdds/dds_api.hpp
+++ b/rmw_connextdds_common/include/rmw_connextdds/dds_api.hpp
@@ -72,7 +72,7 @@ rmw_connextdds_create_contentfilteredtopic(
   DDS_Topic * const base_topic,
   const char * const cft_name,
   const char * const cft_filter,
-  const rcutils_string_array_t * cft_expression_parameters,
+  const rcutils_string_array_t * const cft_expression_parameters,
   DDS_TopicDescription ** const cft_out);
 
 rmw_ret_t
--- a/rmw_connextdds_common/include/rmw_connextdds/dds_api.hpp
+++ b/rmw_connextdds_common/include/rmw_connextdds/dds_api.hpp
@@ -211,7 +211,7 @@ rmw_connextdds_enable_security(
 rmw_ret_t
 rmw_connextdds_set_cft_filter_expression(
   DDS_TopicDescription * const topic_desc,
-  const char * filter_expression,
+  const char * const filter_expression,
   const rcutils_string_array_t * expression_parameters);
 
 rmw_ret_t
--- a/rmw_connextdds_common/include/rmw_connextdds/dds_api.hpp
+++ b/rmw_connextdds_common/include/rmw_connextdds/dds_api.hpp
@@ -218,7 +218,7 @@ rmw_ret_t
 rmw_connextdds_get_cft_filter_expression(
   DDS_TopicDescription * const topic_desc,
   char ** const expr_out,
-  rcutils_string_array_t * cft_params_out);
+  rcutils_string_array_t * const cft_params_out);
 
 // Define some macro aliases for security-related properties
 #ifndef DDS_SECURITY_PROPERTY_PREFIX
--- a/rmw_connextdds_common/include/rmw_connextdds/rmw_impl.hpp
+++ b/rmw_connextdds_common/include/rmw_connextdds/rmw_impl.hpp
@@ -442,7 +442,7 @@ public:
   rmw_ret_t
   set_cft_expression_parameters(
     const char * const filter_expression,
-    const rcutils_string_array_t * expression_parameters
+    const rcutils_string_array_t * const expression_parameters
   );
 
   rmw_ret_t
--- a/rmw_connextdds_common/include/rmw_connextdds/rmw_impl.hpp
+++ b/rmw_connextdds_common/include/rmw_connextdds/rmw_impl.hpp
@@ -448,7 +448,7 @@ public:
   rmw_ret_t
   get_cft_expression_parameters(
     char ** const filter_expression,
-    rcutils_string_array_t * expression_parameters
+    rcutils_string_array_t * const expression_parameters
   );
 
   bool
--- a/rmw_connextdds_common/include/rmw_connextdds/rmw_impl.hpp
+++ b/rmw_connextdds_common/include/rmw_connextdds/rmw_impl.hpp
@@ -353,7 +353,7 @@ public:
   RMW_Connext_SubscriberStatusCondition *
   condition()
   {
-    return &this->status_condition;
+    return this->status_condition.get();
   }
 
   const rmw_gid_t * gid() const
@@ -396,7 +396,7 @@ public:
     }
 
     if (this->internal) {
-      return this->status_condition.trigger_loan_guard_condition(this->loan_len > 0);
+      return this->status_condition->trigger_loan_guard_condition(this->loan_len > 0);
     }
 
     return RMW_RET_OK;
@@ -437,6 +437,18 @@ public:
     rmw_message_info_t * const message_info,
     bool * const taken);
 
+  rmw_ret_t
+  set_cft_expression_parameters(
+    const char * filter_expression,
+    const rcutils_string_array_t * expression_parameters
+  );
+
+  rmw_ret_t
+  get_cft_expression_parameters(
+    char ** filter_expression,
+    rcutils_string_array_t * expression_parameters
+  );
+
   bool
   has_data()
   {
@@ -467,6 +479,21 @@ public:
     return this->dds_topic;
   }
 
+  static std::string get_atomic_id() {
+    static std::atomic_uint64_t id;
+    return std::to_string(id++);
+  }
+
+  void set_node(const rmw_node_t * node)
+  {
+    this->node = node;
+  }
+
+  bool is_cft_supported()
+  {
+    return nullptr != dds_topic_cft;
+  }
+
   const bool internal;
   const bool ignore_local;
 
@@ -478,12 +505,16 @@ private:
   RMW_Connext_MessageTypeSupport * type_support;
   rmw_gid_t ros_gid;
   const bool created_topic;
-  RMW_Connext_SubscriberStatusCondition status_condition;
+  std::shared_ptr<RMW_Connext_SubscriberStatusCondition> status_condition;
   RMW_Connext_UntypedSampleSeq loan_data;
   DDS_SampleInfoSeq loan_info;
-  size_t loan_len;
+  std::atomic_uint64_t loan_len;
   size_t loan_next;
   std::mutex loan_mutex;
+  std::mutex cft_mutex;
+  const rmw_node_t * node;
+  std::string fqtopic_name;
+  rmw_qos_profile_t qos_policies;
 
   RMW_Connext_Subscriber(
     rmw_context_impl_t * const ctx,
--- a/rmw_connextdds_common/include/rmw_connextdds/rmw_impl.hpp
+++ b/rmw_connextdds_common/include/rmw_connextdds/rmw_impl.hpp
@@ -447,7 +447,7 @@ public:
 
   rmw_ret_t
   get_cft_expression_parameters(
-    char ** filter_expression,
+    char ** const filter_expression,
     rcutils_string_array_t * expression_parameters
   );
 
--- a/rmw_connextdds_common/include/rmw_connextdds/rmw_impl.hpp
+++ b/rmw_connextdds_common/include/rmw_connextdds/rmw_impl.hpp
@@ -282,7 +282,7 @@ public:
     const char * const cft_filter = nullptr);
 
   rmw_ret_t
-  finalize();
+  finalize(const bool reset_cft = false);
 
   DDS_DataReader *
   reader() const
@@ -517,6 +517,7 @@ private:
   const rmw_node_t * node;
   std::string fqtopic_name;
   rmw_qos_profile_t qos_policies;
+  rmw_subscription_options_t subscriber_options;
 
   RMW_Connext_Subscriber(
     rmw_context_impl_t * const ctx,
@@ -527,6 +528,19 @@ private:
     const bool created_topic,
     DDS_TopicDescription * const dds_topic_cft,
     const bool internal);
+
+  static
+  DDS_DataReader *
+  initialize_datareader(
+    rmw_context_impl_t * const ctx,
+    DDS_DomainParticipant * const dp,
+    DDS_Subscriber * const sub,
+    const std::string & fqtopic_name,
+    const rmw_qos_profile_t * const qos_policies,
+    RMW_Connext_MessageTypeSupport * const type_support,
+    const rmw_subscription_options_t * const subscriber_options,
+    const bool internal,
+    DDS_TopicDescription * sub_topic);
 
   friend class RMW_Connext_SubscriberStatusCondition;
 };
--- a/rmw_connextdds_common/include/rmw_connextdds/rmw_impl.hpp
+++ b/rmw_connextdds_common/include/rmw_connextdds/rmw_impl.hpp
@@ -441,7 +441,7 @@ public:
 
   rmw_ret_t
   set_cft_expression_parameters(
-    const char * filter_expression,
+    const char * const filter_expression,
     const rcutils_string_array_t * expression_parameters
   );
 
--- a/rmw_connextdds_common/include/rmw_connextdds/rmw_impl.hpp
+++ b/rmw_connextdds_common/include/rmw_connextdds/rmw_impl.hpp
@@ -274,6 +274,7 @@ public:
     const rmw_qos_profile_t * const qos_policies,
     const rmw_subscription_options_t * const subscriber_options,
     const bool internal = false,
+    const rmw_node_t * const node = nullptr,
     const RMW_Connext_MessageType msg_type = RMW_CONNEXT_MESSAGE_USERDATA,
     const void * const intro_members = nullptr,
     const bool intro_members_cpp = false,
@@ -486,11 +487,6 @@ public:
     return std::to_string(id++);
   }
 
-  void set_node(const rmw_node_t * node)
-  {
-    this->node = node;
-  }
-
   bool is_cft_supported()
   {
     return nullptr != dds_topic_cft;
@@ -515,7 +511,6 @@ private:
   std::mutex loan_mutex;
   std::mutex cft_mutex;
   const rmw_node_t * node;
-  std::string fqtopic_name;
   rmw_qos_profile_t qos_policies;
   rmw_subscription_options_t subscriber_options;
 
@@ -527,7 +522,10 @@ private:
     const bool ignore_local,
     const bool created_topic,
     DDS_TopicDescription * const dds_topic_cft,
-    const bool internal);
+    const bool internal,
+    const rmw_node_t * const node,
+    const rmw_qos_profile_t * const qos_policies,
+    const rmw_subscription_options_t * const subscriber_options);
 
   static
   DDS_DataReader *
--- a/rmw_connextdds_common/include/rmw_connextdds/rmw_impl.hpp
+++ b/rmw_connextdds_common/include/rmw_connextdds/rmw_impl.hpp
@@ -510,7 +510,7 @@ private:
   std::shared_ptr<RMW_Connext_SubscriberStatusCondition> status_condition;
   RMW_Connext_UntypedSampleSeq loan_data;
   DDS_SampleInfoSeq loan_info;
-  std::atomic_uint64_t loan_len;
+  size_t loan_len;
   size_t loan_next;
   std::mutex loan_mutex;
   std::mutex cft_mutex;
--- a/rmw_connextdds_common/include/rmw_connextdds/rmw_impl.hpp
+++ b/rmw_connextdds_common/include/rmw_connextdds/rmw_impl.hpp
@@ -510,7 +510,7 @@ private:
   size_t loan_next;
   std::mutex loan_mutex;
   std::mutex cft_mutex;
-  const rmw_node_t * node;
+  const rmw_node_t * const node;
   rmw_qos_profile_t qos_policies;
   rmw_subscription_options_t subscriber_options;
 
--- a/rmw_connextdds_common/include/rmw_connextdds/rmw_impl.hpp
+++ b/rmw_connextdds_common/include/rmw_connextdds/rmw_impl.hpp
@@ -538,7 +538,7 @@ private:
     RMW_Connext_MessageTypeSupport * const type_support,
     const rmw_subscription_options_t * const subscriber_options,
     const bool internal,
-    DDS_TopicDescription * sub_topic);
+    DDS_TopicDescription * const sub_topic);
 
   friend class RMW_Connext_SubscriberStatusCondition;
 };
--- a/rmw_connextdds_common/include/rmw_connextdds/rmw_impl.hpp
+++ b/rmw_connextdds_common/include/rmw_connextdds/rmw_impl.hpp
@@ -18,6 +18,7 @@
 #include <string>
 #include <vector>
 #include <map>
+#include <memory>
 #include <mutex>
 #include <condition_variable>
 #include <atomic>
@@ -479,7 +480,8 @@ public:
     return this->dds_topic;
   }
 
-  static std::string get_atomic_id() {
+  static std::string get_atomic_id()
+  {
     static std::atomic_uint64_t id;
     return std::to_string(id++);
   }
--- a/rmw_connextdds_common/src/common/rmw_impl.cpp
+++ b/rmw_connextdds_common/src/common/rmw_impl.cpp
@@ -1637,7 +1637,7 @@ RMW_Connext_Subscriber::get_cft_expression_parameters(
 {
   std::lock_guard<std::mutex> lock(this->cft_mutex);
   if (nullptr == this->dds_topic_cft) {
-    RMW_CONNEXT_LOG_ERROR_SET("this subscriber has not created a contentfilteredtopic.");
+    RMW_CONNEXT_LOG_ERROR_SET("this subscriber has not created a contentfilteredtopic")
     return RMW_RET_ERROR;
   }
 
--- a/rmw_connextdds_common/src/common/rmw_impl.cpp
+++ b/rmw_connextdds_common/src/common/rmw_impl.cpp
@@ -1637,7 +1637,7 @@ RMW_Connext_Subscriber::get_cft_expression_parameters(
 {
   std::lock_guard<std::mutex> lock(this->cft_mutex);
   if (nullptr == this->dds_topic_cft) {
-    RMW_SET_ERROR_MSG("this subscriber has not created a contentfilteredtopic.");
+    RMW_CONNEXT_LOG_ERROR_SET("this subscriber has not created a contentfilteredtopic.");
     return RMW_RET_ERROR;
   }
 
--- a/rmw_connextdds_common/src/common/rmw_impl.cpp
+++ b/rmw_connextdds_common/src/common/rmw_impl.cpp
@@ -1553,7 +1553,7 @@ RMW_Connext_Subscriber::take_serialized(
 rmw_ret_t
 RMW_Connext_Subscriber::set_cft_expression_parameters(
   const char * const filter_expression,
-  const rcutils_string_array_t * expression_parameters)
+  const rcutils_string_array_t * const expression_parameters)
 {
   RMW_CONNEXT_ASSERT(nullptr != filter_expression)
   RMW_CONNEXT_ASSERT(!this->internal)
--- a/rmw_connextdds_common/src/common/rmw_impl.cpp
+++ b/rmw_connextdds_common/src/common/rmw_impl.cpp
@@ -1216,9 +1216,7 @@ RMW_Connext_Subscriber::create(
 
   DDS_TopicDescription * sub_topic = DDS_Topic_as_topicdescription(topic);
   std::string cft_topic_name;
-  if (nullptr != cft_name ||
-    nullptr != subscriber_options->filter_expression)
-  {
+  if (nullptr != cft_name || nullptr != subscriber_options->filter_expression) {
     rmw_ret_t cft_rc = RMW_RET_OK;
 
     if (nullptr != cft_name) {
@@ -1240,7 +1238,9 @@ RMW_Connext_Subscriber::create(
         return nullptr;
       }
     } else {
-      sub_topic = cft_topic;
+      if (nullptr != cft_topic) {
+        sub_topic = cft_topic;
+      }
     }
   }
 
@@ -1540,26 +1540,23 @@ RMW_Connext_Subscriber::set_cft_expression_parameters(
 {
   RMW_CONNEXT_ASSERT(nullptr != filter_expression)
   RMW_CONNEXT_ASSERT(!internal)
+  RMW_CONNEXT_ASSERT(nullptr != node)
 
   rmw_ret_t ret;
   std::lock_guard<std::mutex> lock(this->cft_mutex);
   const bool filter_expression_empty = (*filter_expression == '\0');
-  if (nullptr == this->dds_topic_cft) {
+  if (nullptr == this->dds_topic_cft && filter_expression_empty) {
     // allow to call set filter_expresson even if cft not exist
-    if (filter_expression_empty) {
-      RMW_CONNEXT_LOG_DEBUG("current subscriber has no content filter topic")
-      return RMW_RET_OK;
-    }
-  } else {
-    if (!filter_expression_empty) {
-      // set filter expre if cft exist
-      return rmw_connextdds_set_cft_filter_expression(
-        this->dds_topic_cft, filter_expression, expression_parameters);
-    }
+    RMW_CONNEXT_LOG_DEBUG("current subscriber has no content filter topic")
+    return RMW_RET_OK;
+  } else if (nullptr != this->dds_topic_cft && !filter_expression_empty) {
+    // set filter expre if cft exist
+    return rmw_connextdds_set_cft_filter_expression(
+      this->dds_topic_cft, filter_expression, expression_parameters);
   }
 
   // finalization to remove the old data reader
-  ret = finalize(true);
+  ret = finalize(true /* reset_cft */);
   if (RMW_RET_OK != ret) {
     RMW_CONNEXT_LOG_ERROR_SET("failed to finalize subscriber with resetting cft flag")
     return ret;
--- a/rmw_connextdds_common/src/common/rmw_impl.cpp
+++ b/rmw_connextdds_common/src/common/rmw_impl.cpp
@@ -1311,7 +1311,7 @@ RMW_Connext_Subscriber::initialize_datareader(
   RMW_Connext_MessageTypeSupport * const type_support,
   const rmw_subscription_options_t * const subscriber_options,
   const bool internal,
-  DDS_TopicDescription * sub_topic)
+  DDS_TopicDescription * const sub_topic)
 {
   // The following initialization generates warnings when built
   // with RTI Connext DDS Professional < 6 (e.g. 5.3.1), so use
--- a/rmw_connextdds_common/src/common/rmw_impl.cpp
+++ b/rmw_connextdds_common/src/common/rmw_impl.cpp
@@ -27,6 +27,7 @@
 const char * const ROS_TOPIC_PREFIX = "rt";
 const char * const ROS_SERVICE_REQUESTER_PREFIX = ROS_SERVICE_REQUESTER_PREFIX_STR;
 const char * const ROS_SERVICE_RESPONSE_PREFIX = ROS_SERVICE_RESPONSE_PREFIX_STR;
+const char * const ROS_CFT_TOPIC_NAME_INFIX = "_ContentFilterTopic";
 
 std::string
 rmw_connextdds_create_topic_name(
@@ -1086,7 +1087,10 @@ RMW_Connext_Subscriber::RMW_Connext_Subscriber(
   const bool ignore_local,
   const bool created_topic,
   DDS_TopicDescription * const dds_topic_cft,
-  const bool internal)
+  const bool internal,
+  const rmw_node_t * const node,
+  const rmw_qos_profile_t * const qos_policies,
+  const rmw_subscription_options_t * const subscriber_options)
 : internal(internal),
   ignore_local(ignore_local),
   ctx(ctx),
@@ -1095,7 +1099,10 @@ RMW_Connext_Subscriber::RMW_Connext_Subscriber(
   dds_topic_cft(dds_topic_cft),
   type_support(type_support),
   created_topic(created_topic),
-  status_condition(new RMW_Connext_SubscriberStatusCondition(dds_reader, ignore_local, internal))
+  status_condition(new RMW_Connext_SubscriberStatusCondition(dds_reader, ignore_local, internal)),
+  node(node),
+  qos_policies(*qos_policies),
+  subscriber_options(*subscriber_options)
 {
   rmw_connextdds_get_entity_gid(this->dds_reader, this->ros_gid);
 
@@ -1124,6 +1131,7 @@ RMW_Connext_Subscriber::create(
   const rmw_qos_profile_t * const qos_policies,
   const rmw_subscription_options_t * const subscriber_options,
   const bool internal,
+  const rmw_node_t * const node,
   const RMW_Connext_MessageType msg_type,
   const void * const intro_members,
   const bool intro_members_cpp,
@@ -1225,7 +1233,7 @@ RMW_Connext_Subscriber::create(
         ctx, dp, topic, cft_name, cft_filter, NULL, &cft_topic);
     } else if (nullptr != subscriber_options->filter_expression) {
       cft_topic_name =
-        fqtopic_name + "_ContentFilterTopic" + RMW_Connext_Subscriber::get_atomic_id();
+        fqtopic_name + ROS_CFT_TOPIC_NAME_INFIX + RMW_Connext_Subscriber::get_atomic_id();
       cft_rc =
         rmw_connextdds_create_contentfilteredtopic(
         ctx, dp, topic, cft_topic_name.c_str(),
@@ -1244,7 +1252,7 @@ RMW_Connext_Subscriber::create(
     }
   }
 
-  DDS_DataReader * dds_reader = initialize_datareader(
+  DDS_DataReader * dds_reader = RMW_Connext_Subscriber::initialize_datareader(
     ctx,
     dp,
     sub,
@@ -1276,15 +1284,15 @@ RMW_Connext_Subscriber::create(
     subscriber_options->ignore_local_publications,
     topic_created,
     cft_topic,
-    internal);
+    internal,
+    node,
+    qos_policies,
+    subscriber_options);
 
   if (nullptr == rmw_sub_impl) {
     RMW_CONNEXT_LOG_ERROR_SET("failed to allocate RMW subscriber")
     return nullptr;
   }
-  rmw_sub_impl->fqtopic_name = fqtopic_name;
-  rmw_sub_impl->qos_policies = *qos_policies;
-  rmw_sub_impl->subscriber_options = *subscriber_options;
 
   scope_exit_type_unregister.cancel();
   scope_exit_topic_delete.cancel();
@@ -1371,17 +1379,24 @@ RMW_Connext_Subscriber::finalize(const bool reset_cft)
     }
   }
 
-  if (nullptr != this->dds_reader && DDS_RETCODE_OK !=
+  if (nullptr == this->dds_reader) {
+    RMW_CONNEXT_LOG_ERROR_SET("DDS DataReader is invalid")
+    return RMW_RET_ERROR;
+  }
+
+  // to get the participant needs a valid dds_reader
+  DDS_DomainParticipant * const participant = this->dds_participant();
+
+  if (DDS_RETCODE_OK !=
     DDS_Subscriber_delete_datareader(
       this->dds_subscriber(), this->dds_reader))
   {
     RMW_CONNEXT_LOG_ERROR_SET("failed to delete DDS DataReader")
     return RMW_RET_ERROR;
   }
+  this->dds_reader = nullptr;
 
   if (!reset_cft) {
-    DDS_DomainParticipant * const participant = this->dds_participant();
-
     if (nullptr != this->dds_topic_cft) {
       rmw_ret_t cft_rc = rmw_connextdds_delete_contentfilteredtopic(
         ctx, participant, this->dds_topic_cft);
@@ -1541,8 +1556,8 @@ RMW_Connext_Subscriber::set_cft_expression_parameters(
   const rcutils_string_array_t * expression_parameters)
 {
   RMW_CONNEXT_ASSERT(nullptr != filter_expression)
-  RMW_CONNEXT_ASSERT(!internal)
-  RMW_CONNEXT_ASSERT(nullptr != node)
+  RMW_CONNEXT_ASSERT(!this->internal)
+  RMW_CONNEXT_ASSERT(nullptr != this->node)
 
   rmw_ret_t ret;
   std::lock_guard<std::mutex> lock(this->cft_mutex);
@@ -1557,21 +1572,25 @@ RMW_Connext_Subscriber::set_cft_expression_parameters(
       this->dds_topic_cft, filter_expression, expression_parameters);
   }
 
+  DDS_DomainParticipant * const dp = this->dds_participant();
+  DDS_Subscriber * const sub = this->dds_subscriber();
+  DDS_TopicDescription * sub_topic = DDS_Topic_as_topicdescription(this->dds_topic);
+  std::string fqtopic_name = DDS_TopicDescription_get_name(sub_topic);
+
   // finalization to remove the old data reader
-  ret = finalize(true /* reset_cft */);
+  ret = this->finalize(true /* reset_cft */);
   if (RMW_RET_OK != ret) {
     RMW_CONNEXT_LOG_ERROR_SET("failed to finalize subscriber with resetting cft flag")
     return ret;
   }
-  DDS_DomainParticipant * const dp = this->dds_participant();
-  DDS_TopicDescription * sub_topic = nullptr;
+
   if (nullptr == this->dds_topic_cft) {
     // create a new cft topic, and then use cft topic to create a new reader
     std::string cft_topic_name =
-      fqtopic_name + "_ContentFilterTopic" + RMW_Connext_Subscriber::get_atomic_id();
+      fqtopic_name + ROS_CFT_TOPIC_NAME_INFIX + RMW_Connext_Subscriber::get_atomic_id();
 
     ret = rmw_connextdds_create_contentfilteredtopic(
-      ctx, dp, dds_topic, cft_topic_name.c_str(),
+      this->ctx, dp, this->dds_topic, cft_topic_name.c_str(),
       filter_expression,
       expression_parameters, &sub_topic);
 
@@ -1584,25 +1603,23 @@ RMW_Connext_Subscriber::set_cft_expression_parameters(
   } else {
     // delete cft, and then use the existing parent topic to create a new reader
     ret = rmw_connextdds_delete_contentfilteredtopic(
-      ctx, dp, this->dds_topic_cft);
+      this->ctx, dp, this->dds_topic_cft);
 
     if (RMW_RET_OK != ret) {
       return ret;
     }
     this->dds_topic_cft = nullptr;
-
-    sub_topic = DDS_Topic_as_topicdescription(dds_topic);
   }
 
-  DDS_DataReader * dds_reader = initialize_datareader(
-    ctx,
+  DDS_DataReader * dds_reader = RMW_Connext_Subscriber::initialize_datareader(
+    this->ctx,
     dp,
-    this->dds_subscriber(),
+    sub,
     fqtopic_name,
-    &qos_policies,
-    type_support,
-    &subscriber_options,
-    internal,
+    &this->qos_policies,
+    this->type_support,
+    &this->subscriber_options,
+    this->internal,
     sub_topic);
 
   if (nullptr == dds_reader) {
@@ -1623,7 +1640,7 @@ RMW_Connext_Subscriber::set_cft_expression_parameters(
     return ret;
   }
 
-  ret = rmw_connextdds_graph_on_subscriber_created(ctx, node, this);
+  ret = rmw_connextdds_graph_on_subscriber_created(this->ctx, this->node, this);
   if (RMW_RET_OK != ret) {
     RMW_CONNEXT_LOG_ERROR("failed to update graph for subscriber")
     return ret;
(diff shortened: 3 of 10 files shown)


fujitatomoya (commented):
overall it seems to be okay. test would be ideal.

rmw_connextdds_common/src/common/rmw_impl.cpp
```diff
@@ -1540,6 +1558,258 @@ RMW_Connext_Subscriber::take_serialized(
...
+    return RMW_RET_ERROR;
+  }
+
+  this->dds_reader = dds_reader;
+
+  this->status_condition.reset(
+    new RMW_Connext_SubscriberStatusCondition(this->dds_reader, this->ignore_local));
+  rmw_connextdds_get_entity_gid(this->dds_reader, this->ros_gid);
+
+  if (!internal) {
```
asorbini, line 1713: In general, I wouldn't expect this function to ever be called on "internal" subscribers (which, at the moment, is only one DataReader created by the RMW for topic `ros_discovery_info`). This check can be removed, and I would instead add a "precondition check" at the beginning of the function to return with an error if the function is called on an "internal" endpoint (it could maybe be an assertion with `RMW_CONNEXT_ASSERT(!internal)`, since there is no way to call it on these endpoints)

asorbini (changes requested):
Things look pretty good.

My main concern is with functions `RMW_Connext_Subscriber::set_cft_expression_parameters`/`get_cft_expression_parameters` which, besides the comments I added inline, have a few other issues that will need to be resolved:

- Function `RMW_Connext_Subscriber::set_cft_expression_parameters` seems to duplicate some logic from `RMW_Connext_Subscriber::finalize` and `RMW_Connext_Subscriber::create`, in order to delete and recreate the underlying `DDS_DataReader`. Instead of duplicating this logic, I would refactor the other two functions and rely on them here.
   - `RMW_Connext_Subscriber::finalize` could be made to take an argument `const bool reset_cft`, which, if `true`, would force the function to perform only a subset of the operations it would usually perform (ie. only: invalidate the status condition wrapper, return loaned samples, delete content filter, delete data reader, and -- not currently done by `finalize()` -- delete the status condition wrapper).
   - Some of the logic in `RMW_Connext_Subscriber::create` could be refactored into a separate `RMW_Connext_Subscriber::initialize_datareader` function to encapsulate the logic required to determine DataReaderQos and create the DataReader. The function might have a parameter (e.g. `const bool on_cft_reset`) to make it behave differently if needed.
   - These are just suggestions, and I'm open to other ways to refactor the code, but I'd like to try to extract the logic to "delete and re-initialize the DataReader" into a separate function, hopefully trying to avoid duplication with the create/finalize functions as much as possible.
- Another problem in these functions is that, as it is, they will not compile with RTI Connext DDS Micro, because this version doesn't support content-filtered topic, thus symbols like `DDS_ContentFilteredTopic` and `DDS_ContentFilteredTopic_get_filter_expression` will not be available.
  - I suggest to introduce at least two new functions, defined in `dds_api.hpp`, and implemented in `dds_api_ndds.cpp` and `dds_api_rtime.cpp` to abstract the logic to get and set the filter expression on a content filtered topic object using only symbols available in both implementations, e.g.:
  - `rmw_ret_t rmw_connextdds_get_cft_filter_expression(DDS_TopicDescription * const topic_desc, char ** const expr_out)`
  - `rmw_ret_t rmw_connextdds_set_cft_filter_expression(DDS_TopicDescription * const topic_desc, struct DDS_StringSeq * const cft_params)`

Once these issues are resolved, I think we will be good to merge (or very very close to it :) )

asorbini:
I just pushed a commit ([598979b](https://github.com/ros2/rmw_connextdds/commit/598979b620460710d4a627a475d103f7034d4238)) which introduced some conflicts with this PR because it refactors the code related to WaitSets and Conditions.

There are now two different implementations of WaitSets and Conditions (which can be selected at compile-time), but as far as this PR is concerned, the only relevant difference I think is a small change in the signature of `RMW_Connext_StatusCondition::RMW_Connext_StatusCondition(...)`.

I apologize for the additional work, but hopefully the conflicts should be easy to fix. Let me know if you have any questions.

iuhilnehc-ynos:
> I just pushed a commit ([598979b](https://github.com/ros2/rmw_connextdds/commit/598979b620460710d4a627a475d103f7034d4238)) which introduced some conflicts with this PR because it refactors the code related to WaitSets and Conditions.
> 
> There are now two different implementations of WaitSets and Conditions (which can be selected at compile-time), but as far as this PR is concerned, the only relevant difference I think is a small change in the signature of `RMW_Connext_StatusCondition::RMW_Connext_StatusCondition(...)`.
> 
> I apologize for the additional work, but hopefully the conflicts should be easy to fix. Let me know if you have any questions.

Thanks for the notification. Even if I fixed all the issues, I think we should not merge this PR, because it leads to a compilation problem since something update about `rmw` is not merged.

rmw_connextdds_common/src/common/rmw_impl.cpp
```diff
@@ -1540,6 +1557,256 @@ RMW_Connext_Subscriber::take_serialized(
   return rc;
 }
 
+rmw_ret_t
+RMW_Connext_Subscriber::set_cft_expression_parameters(
+  const char * filter_expression,
+  const rcutils_string_array_t * expression_parameters)
+{
+  std::lock_guard<std::mutex> lock(this->cft_mutex);
+  if (nullptr == this->dds_topic_cft) {
```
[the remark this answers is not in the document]
iuhilnehc-ynos: I'll keep this branch, otherwise, the [following logic](https://github.com/ros2/rmw_connextdds/blob/bd36713d9f32d9e9396e48a655981bd2c6cae3e4/rmw_connextdds_common/src/common/rmw_impl.cpp#L1601-L1660) need to add such thing,

```
if ((nullptr == this->dds_topic_cft && has_filter_expression) ||
  ((nullptr != this->dds_topic_cft && !has_filter_expression)))
```

rmw_connextdds_common/src/common/rmw_impl.cpp
```diff
@@ -1540,6 +1557,256 @@ RMW_Connext_Subscriber::take_serialized(
   return rc;
 }
 
+rmw_ret_t
+RMW_Connext_Subscriber::set_cft_expression_parameters(
+  const char * filter_expression,
+  const rcutils_string_array_t * expression_parameters)
+{
+  std::lock_guard<std::mutex> lock(this->cft_mutex);
```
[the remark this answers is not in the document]
iuhilnehc-ynos: updated

rmw_connextdds_common/src/common/rmw_impl.cpp
```diff
@@ -1259,11 +1260,25 @@ RMW_Connext_Subscriber::create(
...
 
   DDS_TopicDescription * sub_topic = DDS_Topic_as_topicdescription(topic);
-
-  if (nullptr != cft_name) {
-    rmw_ret_t cft_rc =
-      rmw_connextdds_create_contentfilteredtopic(
-      ctx, dp, topic, cft_name, cft_filter, &cft_topic);
+  std::string cft_topic_name;
+  if (nullptr != cft_name ||
+    nullptr != subscriber_options->filter_expression)
```
asorbini, line 1265: (partially quoting myself from [this comment in another PR](https://github.com/ros2/rmw_connextdds/pull/13#discussion_r601774772))

As you probably saw, this implementation defines a series of "feature flags" in `static_config.hpp` to be able to support multiple versions of ROS2 from the same branch. This is to avoid having to create branches for different versions, which we'll probably end up doing starting with Galactic, but would have been a pain to maintain up until now, while the RMW was in very active development.

For now, I'd like to continue to adhere with this pattern so that the `master` branch remains buildable with Foxy, so I suggest introducing a new flag in `static_config.hpp` to make sure field `filter_expression` is only accessed in the appropriate ROS2 version. Something like:

```cpp
#ifndef RMW_CONNEXT_HAVE_CONTENT_FILTERED_TOPICS
#define RMW_CONNEXT_HAVE_CONTENT_FILTERED_TOPICS \
  (RMW_CONNEXT_RELEASE > RMW_CONNEXT_RELEASE_FOXY)
#endif /* RMW_CONNEXT_HAVE_CONTENT_FILTERED_TOPICS */
```
Then wrap code that accesses `subscriber_options->filter_expression` (here, and below on line 1273) with `#if RMW_CONNEXT_HAVE_CONTENT_FILTERED_TOPICS`.

rmw_connextdds_common/src/common/rmw_impl.cpp
```diff
@@ -1540,6 +1557,256 @@ RMW_Connext_Subscriber::take_serialized(
...
+  } else {
+    if (nullptr != filter_expression && *filter_expression != '\0') {
+      // case 2: normally to set filter expre if cft already support
+      DDS_ContentFilteredTopic * const cft_topic =
+        DDS_ContentFilteredTopic_narrow(dds_topic_cft);
+
+      struct DDS_StringSeq cft_parameters;
+      DDS_StringSeq_initialize(&cft_parameters);
+      if (expression_parameters) {
+        DDS_StringSeq_ensure_length(
```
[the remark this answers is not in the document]
iuhilnehc-ynos: updated

rmw_connextdds_common/src/common/rmw_impl.cpp
```diff
@@ -1540,6 +1557,256 @@ RMW_Connext_Subscriber::take_serialized(
...
+          &cft_parameters, expression_parameters->size, expression_parameters->size);
+        DDS_StringSeq_from_array(&cft_parameters,
```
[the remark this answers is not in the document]
iuhilnehc-ynos: updated

asorbini:
> Thanks for the notification. Even if I fixed all the issues, I think we should not merge this PR, because it leads to a compilation problem since something update about `rmw` is not merged.

I had actually missed it, but yes, this PR can only be merged after changes to `ros2/rmw` have been merged to introduce field `filter_expression` in the subscription options.

Which actually lead me to [request another change](https://github.com/ros2/rmw_connextdds/pull/12#discussion_r601971085) to make sure that the code continues to build with older ROS2 releases.

rmw_connextdds_common/src/common/rmw_impl.cpp
```diff
@@ -1540,6 +1557,256 @@ RMW_Connext_Subscriber::take_serialized(
...
+      return RMW_RET_OK;
+    }
+  } else {
+    if (nullptr != filter_expression && *filter_expression != '\0') {
+      // case 2: normally to set filter expre if cft already support
+      DDS_ContentFilteredTopic * const cft_topic =
+        DDS_ContentFilteredTopic_narrow(dds_topic_cft);
+
+      struct DDS_StringSeq cft_parameters;
+      DDS_StringSeq_initialize(&cft_parameters);
```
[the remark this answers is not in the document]
iuhilnehc-ynos: updated. I used reference for scope exit.

rmw_connextdds_common/src/common/rmw_impl.cpp
```diff
@@ -1540,6 +1557,256 @@ RMW_Connext_Subscriber::take_serialized(
...
+        DDS_StringSeq_ensure_length(
+          &cft_parameters, expression_parameters->size, expression_parameters->size);
+        DDS_StringSeq_from_array(&cft_parameters,
+          const_cast<const char **>(expression_parameters->data),
+          expression_parameters->size);
+      }
+
+      DDS_ReturnCode_t ret =
+        DDS_ContentFilteredTopic_set_expression(cft_topic, filter_expression, &cft_parameters);
+      DDS_StringSeq_finalize(&cft_parameters);
```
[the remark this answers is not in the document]
iuhilnehc-ynos: updated

rmw_connextdds_common/src/ndds/dds_api_ndds.cpp
```diff
@@ -219,20 +219,31 @@ rmw_connextdds_create_contentfilteredtopic(
...
+  const rcutils_string_array_t * cft_expression_parameters,
   DDS_TopicDescription ** const cft_out)
 {
   UNUSED_ARG(ctx);
   RMW_CONNEXT_ASSERT(nullptr != cft_name)
   RMW_CONNEXT_ASSERT(nullptr != cft_filter)
 
   struct DDS_StringSeq cft_parameters = DDS_SEQUENCE_INITIALIZER;
-  DDS_StringSeq_ensure_length(&cft_parameters, 0, 0);
+  DDS_StringSeq_initialize(&cft_parameters);
```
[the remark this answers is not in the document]
iuhilnehc-ynos: updated

rmw_connextdds_common/src/ndds/dds_api_ndds.cpp
```diff
@@ -219,20 +219,31 @@ rmw_connextdds_create_contentfilteredtopic(
...
+  if (cft_expression_parameters) {
+    DDS_StringSeq_ensure_length(
+      &cft_parameters, cft_expression_parameters->size, cft_expression_parameters->size);
+    DDS_StringSeq_from_array(&cft_parameters,
+      const_cast<const char **>(cft_expression_parameters->data),
+      cft_expression_parameters->size);
+  } else {
+    DDS_StringSeq_ensure_length(&cft_parameters, 0, 0);
+  }
```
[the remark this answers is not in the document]
iuhilnehc-ynos: I removed DDS_StringSeq_ensure_length with size 0 which seems not necessary.

rmw_connextdds_common/src/ndds/dds_api_ndds.cpp
```diff
@@ -219,20 +219,31 @@ rmw_connextdds_create_contentfilteredtopic(
...
 
   *cft_out = nullptr;
 
   DDS_ContentFilteredTopic * cft_topic =
     DDS_DomainParticipant_create_contentfilteredtopic(
     dp, cft_name, base_topic, cft_filter, &cft_parameters);
+  DDS_StringSeq_finalize(&cft_parameters);
```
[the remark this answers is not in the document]
iuhilnehc-ynos: updated

rmw_connextdds_common/src/common/rmw_impl.cpp
iuhilnehc-ynos: good to know, thanks.

rmw_connextdds_common/src/common/rmw_impl.cpp
iuhilnehc-ynos: Oh, actually I don't like this style.
I found the `RMW_CONNEXT_HAVE_OPTIONS_PUBSUB` before, which seems a little weird.
But I'll follow your suggestion. :)

iuhilnehc-ynos: It seems you don't want to use some specific branches (eloquent, foxy, etc) to manage the code, but just use one branch(master) with multiple macros.

iuhilnehc-ynos: > This is to avoid having to create branches for different versions, which we'll probably end up doing starting with Galactic

This CFT feature in `rmw` will not be merged in ROS2 Galactic (but might be H version), so I have no idea which way I need to do it?
```
#define RMW_CONNEXT_RELEASE_GALACTIC        50

#ifndef RMW_CONNEXT_HAVE_CONTENT_FILTERED_TOPICS
#define RMW_CONNEXT_HAVE_CONTENT_FILTERED_TOPICS \
  (RMW_CONNEXT_RELEASE > RMW_CONNEXT_RELEASE_GALACTIC)
#endif /* RMW_CONNEXT_HAVE_CONTENT_FILTERED_TOPICS */
```

or 

do nothing. (based on your comment)

iuhilnehc-ynos:
> hopefully the conflicts should be easy to fix. Let me know if you have any questions.

I have rebased from the latest branch of `master`.

rmw_connextdds_common/src/common/rmw_impl.cpp
asorbini: If the feature will not make it into Galactice, I would use `RMW_CONNEXT_RELEASE_ROLLING` for now, i.e.:
```cpp
#ifndef RMW_CONNEXT_HAVE_CONTENT_FILTERED_TOPICS
#define RMW_CONNEXT_HAVE_CONTENT_FILTERED_TOPICS \
    (RMW_CONNEXT_RELEASE >= RMW_CONNEXT_RELEASE_ROLLING)
#endif /* RMW_CONNEXT_HAVE_CONTENT_FILTERED_TOPICS */
```

I will review all feature flags, introduce `RMW_CONNEXT_RELEASE_GALACTIC`, and bump the value of `RMW_CONNEXT_RELEASE_ROLLING` after this feature is merged.

I know having macros is not the best, but it was the "lesser evil" as the code base was in very active development to allow supporting multiple version without having to deal with merging a lot between branches.

Once the package is included in a release, I think we will create a dedicated `galactic` branch for maintenance purposes (only bug fixes and critical patches will be merged). We will also be able to use that branch for releases pre-Galactic (along with the existing `dashing` branch), and at that point we can consider cleaning up the main development branch (`master`) to only support Rolling and future releases.

rmw_connextdds_common/src/common/rmw_impl.cpp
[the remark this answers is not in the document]
asorbini: I see, sorry, I somehow ignored the `return`, my bad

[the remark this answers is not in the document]
asorbini: still, it feels a little weird to me... I think I would prefer:

```cpp
if (nullptr == this->dds_topic_cft && filter_expression_empty) {
  ...
  return RMW_RET_OK;
} else if if (nullptr != this->dds_topic_cft) {
  ...
}
```

rmw_connextdds_common/src/common/rmw_impl.cpp
```diff
@@ -1530,6 +1579,123 @@ RMW_Connext_Subscriber::take_serialized(
...
+  } else {
+    if (!filter_expression_empty) {
+      // set filter expre if cft exist
+      return rmw_connextdds_set_cft_filter_expression(
+        this->dds_topic_cft, filter_expression, expression_parameters);
+    }
+  }
+
+  // finalization to remove the old data reader
+  ret = finalize(true);
```
asorbini, line 1608: Please add a comment with the parameter name, so it's easier to read the code and know what the literal value is used for (e.g. `true /* reset_cft */`).

rmw_connextdds_common/src/ndds/dds_api_ndds.cpp
```diff
@@ -212,21 +212,56 @@ rmw_connextdds_initialize_participant_qos_impl(
   return RMW_RET_OK;
 }
 
+static
+rmw_ret_t
+initialize_cft_parameters(
```
asorbini, line 217: for consistency with other symbols, please add prefix `rmw_connextdds_` to function name.

rmw_connextdds_common/src/common/rmw_impl.cpp
```diff
@@ -1366,52 +1413,54 @@ RMW_Connext_Subscriber::finalize()
...
+      this->ctx, participant, this->type_support->type_name());
 
-  delete this->type_support;
-  this->type_support = nullptr;
+    if (RMW_RET_OK != rc) {
+      return rc;
+    }
 
+    delete this->type_support;
+    this->type_support = nullptr;
```
asorbini, line 1462: The type support shouldn't be deleted if `reset_cft == true`. Extend the block above to include these lines.

rmw_connextdds_common/src/common/rmw_impl.cpp
```diff
@@ -1530,6 +1579,123 @@ RMW_Connext_Subscriber::take_serialized(
   return rc;
 }
 
+rmw_ret_t
+RMW_Connext_Subscriber::set_cft_expression_parameters(
+  const char * filter_expression,
+  const rcutils_string_array_t * expression_parameters)
+{
+  RMW_CONNEXT_ASSERT(nullptr != filter_expression)
+  RMW_CONNEXT_ASSERT(!internal)
```
asorbini, line 1588: Just to be safe, let's add an assertion for `RMW_CONNEXT_ASSERT(nullptr != node)`, just to keep track that that's a prerequisite too since `node` might be null in some cases. For now this only happens for `internal` endpoints, so the two assertions kind of overlap, but maybe in the future things might deviate and we might forget of the other prerequisite of this function, so the explicit reminder will help.

rmw_connextdds_common/src/common/rmw_impl.cpp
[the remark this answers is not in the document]
asorbini: didn't know I could use that syntax to capture the variable as a reference, thank you for teaching me something!

rmw_connextdds_common/src/ndds/dds_api_ndds.cpp
```diff
@@ -1275,3 +1310,113 @@ rmw_connextdds_enable_security(
...
+    [expr_out, allocator]()
+    {
+      if (*expr_out) {
+        allocator.deallocate(*expr_out, allocator.state);
+        *expr_out = nullptr;
+      }
+    });
+
+  // get parameters
+  struct DDS_StringSeq parameters;
```
asorbini, line 1381: Initialize sequence with `= DDS_SEQUENCE_INITIALIZER;`

rmw_connextdds_common/src/ndds/dds_api_ndds.cpp
```diff
@@ -1275,3 +1310,113 @@ rmw_connextdds_enable_security(
...
+  auto scope_exit_expression_parameters_delete =
+    rcpputils::make_scope_exit(
+    [cft_params_out]()
+    {
+      if (RCUTILS_RET_OK != rcutils_string_array_fini(cft_params_out)) {
+        RCUTILS_LOG_ERROR("Error while finalizing expression parameter due to another error");
+      }
+    });
+  for (int i = 0; i < parameters_len; ++i) {
+    char * parameter = rcutils_strdup(DDS_StringSeq_get(&parameters, i), allocator);
```
asorbini, line 1411: Please put call to `DDS_StringSeq_get()` on a separate line and cache result to a local variable. Also, although they are equivalent, please use `*DDS_StringSeq_get_reference(&parameters, i)`, since the `get_reference()` function is used for all sequence access in the rest of the RMW. Thanks!

rmw_connextdds_common/src/common/rmw_impl.cpp
iuhilnehc-ynos: the `type_support` block is already extented into the `if(!reset_cft)` branch

rmw_connextdds_common/src/common/rmw_impl.cpp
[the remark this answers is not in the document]
iuhilnehc-ynos: updated

rmw_connextdds_common/src/common/rmw_impl.cpp
iuhilnehc-ynos: thank you. updated

rmw_connextdds_common/src/common/rmw_impl.cpp
iuhilnehc-ynos: updated

rmw_connextdds_common/src/ndds/dds_api_ndds.cpp
iuhilnehc-ynos: updated

rmw_connextdds_common/src/common/rmw_impl.cpp
iuhilnehc-ynos: updated

rmw_connextdds_common/src/ndds/dds_api_ndds.cpp
iuhilnehc-ynos: updated

rmw_connextdds_common/src/ndds/dds_api_ndds.cpp
iuhilnehc-ynos: updated

rmw_connextdds_common/src/common/rmw_impl.cpp
asorbini: sorry, my bad

rmw_connextdds_common/src/ndds/dds_api_ndds.cpp
```diff
@@ -1275,3 +1312,120 @@ rmw_connextdds_enable_security(
...
+  // get filter_expression
+  const char * expression = DDS_ContentFilteredTopic_get_filter_expression(cft_topic);
+  if (!expression) {
+    RMW_SET_ERROR_MSG("failed to get filter expression");
+    return RMW_RET_ERROR;
+  }
+
+  *expr_out = rcutils_strdup(expression, allocator);
+  if (NULL == *expr_out) {
+    RMW_SET_ERROR_MSG("failed to duplicate string");
```
asorbini, line 1370: Sorry, but I hadn't noticed you used these different macros here.

Please replace `RMW_SET_ERROR_MSG(...)` with `RMW_CONNEXT_LOG_ERROR_SET(...)`, and `RMW_SET_ERROR_MSG_WITH_FORMAT_STRING(...)` with `RMW_CONNEXT_LOG_ERROR_A_SET(...)`.

You should also remove the `;` from the log lines after switching macros.

rmw_connextdds_common/src/common/rmw_impl.cpp
asorbini: great, thanks!

asorbini:
Things are looking good, thank you @iuhilnehc-ynos for addressing all my many comments! (only one left)

rmw_connextdds_common/src/ndds/dds_api_ndds.cpp
iuhilnehc-ynos: Don't do that, I think I should say 'sorry'. Thanks for your review, it's very nice of you.

asorbini: 🙏🏼 It's my pleasure to help you, I'm very glad to have others interested in this software!

rmw_connextdds_common/include/rmw_connextdds/rmw_impl.hpp
```diff
@@ -559,12 +588,17 @@ class RMW_Connext_Subscriber
   RMW_Connext_MessageTypeSupport * type_support;
   rmw_gid_t ros_gid;
   const bool created_topic;
-  RMW_Connext_SubscriberStatusCondition status_condition;
+  std::shared_ptr<RMW_Connext_SubscriberStatusCondition> status_condition;
   RMW_Connext_UntypedSampleSeq loan_data;
   DDS_SampleInfoSeq loan_info;
-  size_t loan_len;
+  std::atomic_uint64_t loan_len;
```
asorbini, line 594: What is the reason for making this a `std::atomic_uint64_t`? Access to `loan_len` and other loan-related fields is already protected by `std::mutex loan_mutex`.

iuhilnehc-ynos: I think https://github.com/ros2/rmw_connextdds/blob/3e4fb9067b7bab7f7d8180ac3ed3c3054fa9933d/rmw_connextdds_common/src/common/rmw_impl.cpp#L1419-L1424 should be protected. I'll not use std::atomic_uint64 but add `std::mutex loan_mutex` for it. What do you think?

asorbini: Well, access needs to be synchronized only if applications are allowed to change the CFT settings while concurrently taking messages from the subscription (the call to `status_condition->invalidate()` should already guarantee that the subscription is not attached to any waitset, which might also end up updating the "loan state").

Before these changes, the code in `RMW_Connext_Subscriber::finalize()` assumed the operation wasn't "thread-safe" with respect to other subscription operations.

If this is not the case (i.e. an application is allowed to modify the subscription's CFT parameters while also polling the subscription for samples), then access to `loan_len` should be guarded inside `finalize()`.

By the way, I think the line `this->loan_next = this->loan_len;` can be safely removed either way (I'm not sure why it's there after reviewing the code...). Would you mind making that change?

asorbini:
I created #15 to get rid of all "feature flags" in `master`. If you want, I think you could rebase your changes on `asorbini/rolling-no-ff` (or rebase them from `master`, once that PR is merged).

rmw_connextdds_common/include/rmw_connextdds/rmw_impl.hpp
asorbini: > 
> By the way, I think the line `this->loan_next = this->loan_len;` can be safely removed either way (I'm not sure why it's there after reviewing the code...). Would you mind making that change?

Actually, I think my intention there was to "consume all remaining samples". Still, it doesn't actually do that (or rather, there is no need to do it since the field is immediately reset by `return_messages()`...), so I would still get rid of the line.

iuhilnehc-ynos: OK, I'll update it.

iuhilnehc-ynos:
> rebase them from `master`, once that PR is merged

I'll wait to rebase from master until that PR is merged because there might exist other updates in the future.

asorbini:
Latest changes look good, holding off on approval until we can rebase on top of `master` after merging #15.

asorbini:
#15 was merged, so you can rebase from `master` now and remove the feature flags

iuhilnehc-ynos:
> #15 was merged, so you can rebase from `master` now and remove the feature flags

Thank you for your detailed instructions.

rmw_connextdds_common/src/common/rmw_impl.cpp
```diff
@@ -1730,7 +1889,7 @@ rmw_connextdds_create_subscriber(
       "failed to allocate RMW_Connext_Subscriber")
     return nullptr;
   }
-
+  rmw_sub_impl->set_node(node);
```
asorbini, line 1892: After some further consideration, I would remove `RMW_Connext_Subscriber::set_node()` and instead add a `node` argument to both `RMW_Connext_Subscriber::create()` and `RMW_Connext_Subscriber::RMW_Connext_Subscriber()`.

The only reason, I'm assuming, why it wasn't added as a parameter in the first place is because those functions are also used to create "internal" subscribers (e.g. in `rmw_graph.cpp`), which do not belong to any node.

I think it's more consistent to make the "pointer to the parent node" a creation argument, to make it clear that it's something that it's not going to change over the lifetime of the subscriber, and pass `nullptr` for it when creating an `internal` endpoint, rather than adding a setter for the field, which suggests "mutability", even though the field is not expected to change.

I suggest adding a `const rmw_node_t * const node` parameter to `RMW_Connext_Subscriber::create()` and `RMW_Connext_Subscriber::RMW_Connext_Subscriber()`, and then set `node` in the initialization list in `RMW_Connext_Subscriber::RMW_Connext_Subscriber()`.

rmw_connextdds_common/src/common/rmw_impl.cpp
```diff
@@ -1488,6 +1535,118 @@ RMW_Connext_Subscriber::take_serialized(
...
+
+    if (RMW_RET_OK != ret) {
+      return ret;
+    }
+    this->dds_topic_cft = nullptr;
+
+    sub_topic = DDS_Topic_as_topicdescription(dds_topic);
+  }
+
+  DDS_DataReader * dds_reader = initialize_datareader(
```
asorbini, line 1597: Please qualify the arguments to this function with `this->` to make it clear that we are accessing instance members/methods.

This is of course a "coding style" change to make the code consistent, more than a functional one. I'm not sure if qualifying every access with `this->` is the "modern C++ programmer's way", but I've been consistent so far in doing so (because I personally think it makes code much more readable). We should be consistent with that strategy (as flawed as it may be) until the moment we decide to not do it any more, and update all the code base accordingly.

I will try to point out some other occurrences of this, but please take a pass at the code to make it consistent with this rule.

rmw_connextdds_common/src/common/rmw_impl.cpp
```diff
@@ -1488,6 +1535,118 @@ RMW_Connext_Subscriber::take_serialized(
...
+  if (RMW_RET_OK != ret) {
+    RMW_CONNEXT_LOG_ERROR_SET("failed to finalize subscriber with resetting cft flag")
+    return ret;
+  }
+  DDS_DomainParticipant * const dp = this->dds_participant();
+  DDS_TopicDescription * sub_topic = nullptr;
+  if (nullptr == this->dds_topic_cft) {
+    // create a new cft topic, and then use cft topic to create a new reader
+    std::string cft_topic_name =
+      fqtopic_name + "_ContentFilterTopic" + RMW_Connext_Subscriber::get_atomic_id();
```
asorbini, line 1571: `this->fqtopic_name`

rmw_connextdds_common/src/common/rmw_impl.cpp
```diff
@@ -1488,6 +1535,118 @@ RMW_Connext_Subscriber::take_serialized(
...
+
+    ret = rmw_connextdds_create_contentfilteredtopic(
+      ctx, dp, dds_topic, cft_topic_name.c_str(),
```
asorbini, line 1574: `this->ctx`, ..., `this->dds_topic`

rmw_connextdds_common/src/common/rmw_impl.cpp
```diff
@@ -1488,6 +1535,118 @@ RMW_Connext_Subscriber::take_serialized(
...
+    if (RMW_RET_OK != ret) {
+      RMW_CONNEXT_LOG_ERROR("failed to create content filter topic")
+      return ret;
+    }
+
+    this->dds_topic_cft = sub_topic;
+  } else {
+    // delete cft, and then use the existing parent topic to create a new reader
+    ret = rmw_connextdds_delete_contentfilteredtopic(
+      ctx, dp, this->dds_topic_cft);
```
asorbini, line 1587: `this->ctx`

rmw_connextdds_common/src/common/rmw_impl.cpp
```diff
@@ -1488,6 +1535,118 @@ RMW_Connext_Subscriber::take_serialized(
   return rc;
 }
 
+rmw_ret_t
+RMW_Connext_Subscriber::set_cft_expression_parameters(
+  const char * filter_expression,
+  const rcutils_string_array_t * expression_parameters)
+{
+  RMW_CONNEXT_ASSERT(nullptr != filter_expression)
+  RMW_CONNEXT_ASSERT(!internal)
```
asorbini, line 1544: `this->internal`

rmw_connextdds_common/src/common/rmw_impl.cpp
```diff
@@ -1488,6 +1535,118 @@ RMW_Connext_Subscriber::take_serialized(
...
+  RMW_CONNEXT_ASSERT(nullptr != node)
```
asorbini, line 1545: `this->node`

rmw_connextdds_common/src/common/rmw_impl.cpp
```diff
@@ -1488,6 +1535,118 @@ RMW_Connext_Subscriber::take_serialized(
...
+    RMW_CONNEXT_LOG_DEBUG("current subscriber has no content filter topic")
+    return RMW_RET_OK;
+  } else if (nullptr != this->dds_topic_cft && !filter_expression_empty) {
+    // set filter expre if cft exist
+    return rmw_connextdds_set_cft_filter_expression(
+      this->dds_topic_cft, filter_expression, expression_parameters);
+  }
+
+  // finalization to remove the old data reader
+  ret = finalize(true /* reset_cft */);
```
asorbini, line 1561: `this->finalize()`

rmw_connextdds_common/src/common/rmw_impl.cpp
```diff
@@ -1488,6 +1535,118 @@ RMW_Connext_Subscriber::take_serialized(
...
+      this->dds_reader, this->ignore_local, this->internal));
+  rmw_connextdds_get_entity_gid(this->dds_reader, this->ros_gid);
+
+  ret = this->enable();
+  if (RMW_RET_OK != ret) {
+    RMW_CONNEXT_LOG_ERROR("failed to enable subscription")
+    return ret;
+  }
+
+  ret = rmw_connextdds_graph_on_subscriber_created(ctx, node, this);
```
asorbini, line 1626: `this->ctx`, `this->node`

rmw_connextdds_common/src/common/rmw_impl.cpp
```diff
@@ -1274,104 +1349,76 @@ RMW_Connext_Subscriber::create(
...
-  DDS_DomainParticipant * const participant = this->dds_participant();
+  if (!reset_cft) {
+    DDS_DomainParticipant * const participant = this->dds_participant();
 
-  if (nullptr != this->dds_topic_cft) {
-    rmw_ret_t cft_rc = rmw_connextdds_delete_contentfilteredtopic(
-      ctx, participant, this->dds_topic_cft);
+    if (nullptr != this->dds_topic_cft) {
+      rmw_ret_t cft_rc = rmw_connextdds_delete_contentfilteredtopic(
+        ctx, participant, this->dds_topic_cft);
```
asorbini, line 1387: `this->ctx`

rmw_connextdds_common/src/common/rmw_impl.cpp
```diff
@@ -1274,104 +1349,76 @@ RMW_Connext_Subscriber::create(
...
   }
 
-  if (DDS_RETCODE_OK !=
+  if (nullptr != this->dds_reader && DDS_RETCODE_OK !=
     DDS_Subscriber_delete_datareader(
       this->dds_subscriber(), this->dds_reader))
   {
     RMW_CONNEXT_LOG_ERROR_SET("failed to delete DDS DataReader")
     return RMW_RET_ERROR;
   }
```
asorbini, line 1380: This block fails to reset `this->dds_reader`, this is a problem if we assume `finalize()` is idempotent.

I suggest refactoring to:

```cpp
if (nullptr != this->dds_reader) {
  if (DDS_RETCODE_OK !=
    DDS_Subscriber_delete_datareader(this->dds_subscriber(), this->dds_reader))
  {
    RMW_CONNEXT_LOG_ERROR_SET("failed to delete DDS DataReader")
    return RMW_RET_ERROR;
  }
  this->dds_reader = nullptr;
}
```

You could also just always reset `this->dds_reader`, but I think the nested block is more readable and more consistent with the rest of the the code.

rmw_connextdds_common/src/common/rmw_impl.cpp
```diff
@@ -1215,21 +1215,96 @@ RMW_Connext_Subscriber::create(
...
       }
     } else {
-      sub_topic = cft_topic;
+      if (nullptr != cft_topic) {
+        sub_topic = cft_topic;
+      }
     }
   }
 
+  DDS_DataReader * dds_reader = initialize_datareader(
```
asorbini, line 1247: `RMW_Connext_Subscriber::initialize_datareader(`

rmw_connextdds_common/src/common/rmw_impl.cpp
asorbini: Also, qualify the function call as `RMW_Connext_Subscriber::initialize_datareader(`

rmw_connextdds_common/src/common/rmw_impl.cpp
```diff
@@ -1215,21 +1215,96 @@ RMW_Connext_Subscriber::create(
...
+    subscriber_options->ignore_local_publications,
+    topic_created,
+    cft_topic,
+    internal);
+
+  if (nullptr == rmw_sub_impl) {
+    RMW_CONNEXT_LOG_ERROR_SET("failed to allocate RMW subscriber")
+    return nullptr;
+  }
+  rmw_sub_impl->fqtopic_name = fqtopic_name;
```
asorbini, line 1285: There is no need to cache the "fully qualified" topic name, since that is already available from the `DDS_Topic` object.

When needed, you can access the value as a C string with:

```cpp
const char * const fqtopic_str = DDS_TopicDescription_get_name(DDS_Topic_as_topicdescription(this->dds_topic));
```

If needed, you can construct an `std::string` from that, and even if that ends up doing some memory alloc/copy, this shouldn't be a problem since AFAICT `fqtopic_name` in only accessed inside `MW_Connext_Subscriber::set_cft_expression_parameters()` which isn't a "critical path" function anyway, since it already does (potentially) plenty of allocation/deallocation (particularty when accessing that field).

rmw_connextdds_common/src/common/rmw_impl.cpp
```diff
@@ -1215,21 +1215,96 @@ RMW_Connext_Subscriber::create(
...
+  rmw_sub_impl->qos_policies = *qos_policies;
+  rmw_sub_impl->subscriber_options = *subscriber_options;
```
asorbini, line 1287: I would prefer if these fields were passed directly to the constructor, and cached inside, for considerations similar to the one I mentioned about the `node` fields (making it clear that these are values determined at creation time and which will remain unchanged for the duration of the object's lifetime).

It will also help us keep track of them if they ever develop from "simple data" which doesn't require explicit finalization, to some more complex structure with their own `finalize()` function to call.

asorbini (changes requested):
Thank you for the latest updates @iuhilnehc-ynos.

I left a few more comments about making the code "consistent" with the current coding style of the repository, and a couple of changes to make it (imho) a bit more "future proof".

rmw_connextdds_common/src/common/rmw_impl.cpp
asorbini: Calls to `RMW_Connext_Subscriber::create()` which do not have a `node` available (e.g. in `rmw_graph.cpp`) should pass `nullptr /* node */`.

rmw_connextdds_common/src/common/rmw_impl.cpp
iuhilnehc-ynos: Putting `this->dds_reader = nullptr;` after `DDS_Subscriber_delete_datareader` seems reasonable.

rmw_connextdds_common/src/common/rmw_impl.cpp
iuhilnehc-ynos: Updated

rmw_connextdds_common/src/common/rmw_impl.cpp
```diff
@@ -1371,17 +1379,24 @@ RMW_Connext_Subscriber::finalize(const bool reset_cft)
     }
   }
 
-  if (nullptr != this->dds_reader && DDS_RETCODE_OK !=
+  if (nullptr == this->dds_reader) {
```
asorbini, line 1382: I would move this check before the previous block, because logically "having a DDS data reader" is also a precondition for calling `this->return_messages()`, not just for `this->dds_participant()` (good catch on that).

Also, maybe change the error message to something a bit more specific, like "no DDS DataReader to finalize".

asorbini:
Looks great @iuhilnehc-ynos, thanks!

A few things remaining:
- Missed [one update for `this->`](https://github.com/ros2/rmw_connextdds/pull/12#discussion_r604337427).
- I'm not sure if you want to [remove `fqtopic_name`](https://github.com/ros2/rmw_connextdds/pull/12#discussion_r604344723).
- Minor, but I would [move the check for `this->dds_reader` in `finalize()`](https://github.com/ros2/rmw_connextdds/pull/12#discussion_r605116675).

After that, I think I'm ready to stop nitpicking since the code is in great shape :)

rmw_connextdds_common/src/common/rmw_impl.cpp
asorbini, line 1402:
```suggestion
        this->ctx, participant, this->dds_topic_cft);
```

iuhilnehc-ynos:
> Looks great @iuhilnehc-ynos, thanks!
> 
> A few things remaining:
> 
> * Missed [one update for `this->`](https://github.com/ros2/rmw_connextdds/pull/12#discussion_r604337427).

Ah, even if I checked twice, I still missed one :(

> * ~I'm not sure if you want to [remove `fqtopic_name`](https://github.com/ros2/rmw_connextdds/pull/12#discussion_r604344723).~ I accidentally didn't review the correct set of changes, I see now that it has been fixed.

Removing `fqtopic_name` and then using `DDS_TopicDescription_get_name` seems good, is this a problem?

> * Minor, but I would [move the check for `this->dds_reader` in `finalize()`](https://github.com/ros2/rmw_connextdds/pull/12#discussion_r605116675).

Ennn, seem reasonable.

asorbini:
> Ah, even if I checked twice, I still missed one :(
> 

No problem, it happens :) You can just accept my suggestion if you want.

> 
> Removing `fqtopic_name` and then using `DDS_TopicDescription_get_name` seems good, is this a problem?

It is good. What I meant was to disregard my comment about still having to remove `fqtopic_name`, because I later realized that you had already removed it from `RMW_Connext_Subscriber` with the latest commits.

> 
> > * Minor, but I would [move the check for `this->dds_reader` in `finalize()`](https://github.com/ros2/rmw_connextdds/pull/12#discussion_r605116675).
> 
> Ennn, seem reasonable.

👍🏼

rmw_connextdds_common/include/rmw_connextdds/dds_api.hpp
```diff
@@ -72,6 +72,7 @@ rmw_connextdds_create_contentfilteredtopic(
   DDS_Topic * const base_topic,
   const char * const cft_name,
   const char * const cft_filter,
+  const rcutils_string_array_t * cft_expression_parameters,
```
asorbini, line 75:
```suggestion
  const rcutils_string_array_t * const cft_expression_parameters,
```

rmw_connextdds_common/include/rmw_connextdds/dds_api.hpp
```diff
@@ -207,6 +208,18 @@ rmw_connextdds_enable_security(
   rmw_context_impl_t * const ctx,
   DDS_DomainParticipantQos * const qos);
 
+rmw_ret_t
+rmw_connextdds_set_cft_filter_expression(
+  DDS_TopicDescription * const topic_desc,
+  const char * filter_expression,
```
asorbini, line 214:
```suggestion
  const char * const filter_expression,
```

rmw_connextdds_common/include/rmw_connextdds/dds_api.hpp
```diff
@@ -207,6 +208,18 @@ rmw_connextdds_enable_security(
...
+  const rcutils_string_array_t * expression_parameters);
```
asorbini, line 215:
```suggestion
  const rcutils_string_array_t * const expression_parameters);
```

rmw_connextdds_common/include/rmw_connextdds/dds_api.hpp
```diff
@@ -207,6 +208,18 @@ rmw_connextdds_enable_security(
...
+
+rmw_ret_t
+rmw_connextdds_get_cft_filter_expression(
+  DDS_TopicDescription * const topic_desc,
+  char ** const expr_out,
+  rcutils_string_array_t * cft_params_out);
```
asorbini, line 221:
```suggestion
  rcutils_string_array_t * const cft_params_out);
```

rmw_connextdds_common/include/rmw_connextdds/rmw_impl.hpp
```diff
@@ -437,6 +439,18 @@ class RMW_Connext_Subscriber
     rmw_message_info_t * const message_info,
     bool * const taken);
 
+  rmw_ret_t
+  set_cft_expression_parameters(
+    const char * filter_expression,
```
asorbini, line 444:
```suggestion
    const char * const filter_expression,
```

rmw_connextdds_common/include/rmw_connextdds/rmw_impl.hpp
```diff
@@ -437,6 +439,18 @@ class RMW_Connext_Subscriber
...
+    const rcutils_string_array_t * expression_parameters
```
asorbini, line 445:
```suggestion
    const rcutils_string_array_t * const expression_parameters
```

rmw_connextdds_common/include/rmw_connextdds/rmw_impl.hpp
```diff
@@ -437,6 +439,18 @@ class RMW_Connext_Subscriber
...
+  );
+
+  rmw_ret_t
+  get_cft_expression_parameters(
+    char ** filter_expression,
```
asorbini, line 450:
```suggestion
    char ** const filter_expression,
```

rmw_connextdds_common/include/rmw_connextdds/rmw_impl.hpp
```diff
@@ -437,6 +439,18 @@ class RMW_Connext_Subscriber
...
+    rcutils_string_array_t * expression_parameters
```
asorbini, line 451:
```suggestion
    rcutils_string_array_t * const expression_parameters
```

rmw_connextdds_common/include/rmw_connextdds/rmw_impl.hpp
```diff
@@ -478,12 +503,16 @@ class RMW_Connext_Subscriber
...
   const bool created_topic;
-  RMW_Connext_SubscriberStatusCondition status_condition;
+  std::shared_ptr<RMW_Connext_SubscriberStatusCondition> status_condition;
   RMW_Connext_UntypedSampleSeq loan_data;
   DDS_SampleInfoSeq loan_info;
   size_t loan_len;
   size_t loan_next;
   std::mutex loan_mutex;
+  std::mutex cft_mutex;
+  const rmw_node_t * node;
```
asorbini, line 513:
```suggestion
  const rmw_node_t * const node;
```

rmw_connextdds_common/include/rmw_connextdds/rmw_impl.hpp
```diff
@@ -493,7 +522,23 @@ class RMW_Connext_Subscriber
...
+  initialize_datareader(
+    rmw_context_impl_t * const ctx,
+    DDS_DomainParticipant * const dp,
+    DDS_Subscriber * const sub,
+    const std::string & fqtopic_name,
+    const rmw_qos_profile_t * const qos_policies,
+    RMW_Connext_MessageTypeSupport * const type_support,
+    const rmw_subscription_options_t * const subscriber_options,
+    const bool internal,
+    DDS_TopicDescription * sub_topic);
```
asorbini, line 541:
```suggestion
    DDS_TopicDescription * const sub_topic);
```

rmw_connextdds_common/src/common/rmw_impl.cpp
```diff
@@ -1215,21 +1223,96 @@ RMW_Connext_Subscriber::create(
...
+RMW_Connext_Subscriber::initialize_datareader(
+  rmw_context_impl_t * const ctx,
+  DDS_DomainParticipant * const dp,
+  DDS_Subscriber * const sub,
+  const std::string & fqtopic_name,
+  const rmw_qos_profile_t * const qos_policies,
+  RMW_Connext_MessageTypeSupport * const type_support,
+  const rmw_subscription_options_t * const subscriber_options,
+  const bool internal,
+  DDS_TopicDescription * sub_topic)
```
asorbini, line 1314:
```suggestion
  DDS_TopicDescription * const sub_topic)
```

rmw_connextdds_common/src/common/rmw_impl.cpp
```diff
@@ -1488,6 +1550,120 @@ RMW_Connext_Subscriber::take_serialized(
   return rc;
 }
 
+rmw_ret_t
+RMW_Connext_Subscriber::set_cft_expression_parameters(
+  const char * filter_expression,
```
asorbini, line 1555:
```suggestion
  const char * const filter_expression,
```

rmw_connextdds_common/src/common/rmw_impl.cpp
```diff
@@ -1488,6 +1550,120 @@ RMW_Connext_Subscriber::take_serialized(
...
+  const rcutils_string_array_t * expression_parameters)
```
asorbini, line 1556:
```suggestion
  const rcutils_string_array_t * const expression_parameters)
```

rmw_connextdds_common/src/common/rmw_impl.cpp
```diff
@@ -1488,6 +1550,120 @@ RMW_Connext_Subscriber::take_serialized(
...
+    RMW_CONNEXT_LOG_ERROR("failed to update graph for subscriber")
+    return ret;
+  }
+
+  return RMW_RET_OK;
+}
+
+rmw_ret_t
+RMW_Connext_Subscriber::get_cft_expression_parameters(
+  char ** filter_expression,
```
asorbini, line 1654:
```suggestion
  char ** const filter_expression,
```

rmw_connextdds_common/src/common/rmw_impl.cpp
```diff
@@ -1488,6 +1550,120 @@ RMW_Connext_Subscriber::take_serialized(
...
+  rcutils_string_array_t * expression_parameters)
```
asorbini, line 1655:
```suggestion
  rcutils_string_array_t * const expression_parameters)
```

rmw_connextdds_common/src/ndds/dds_api_ndds.cpp
```diff
@@ -210,21 +210,58 @@ rmw_connextdds_initialize_participant_qos_impl(
   return RMW_RET_OK;
 }
 
+static
+rmw_ret_t
+rmw_connextdds_initialize_cft_parameters(
+  struct DDS_StringSeq * cft_parameters,
```
asorbini, line 216:
```suggestion
  struct DDS_StringSeq * const cft_parameters,
```

rmw_connextdds_common/src/ndds/dds_api_ndds.cpp
```diff
@@ -210,21 +210,58 @@ rmw_connextdds_initialize_participant_qos_impl(
...
+  const rcutils_string_array_t * cft_expression_parameters)
```
asorbini, line 217:
```suggestion
  const rcutils_string_array_t * const cft_expression_parameters)
```

rmw_connextdds_common/src/ndds/dds_api_ndds.cpp
```diff
@@ -210,21 +210,58 @@ rmw_connextdds_initialize_participant_qos_impl(
...
+}
+
 rmw_ret_t
 rmw_connextdds_create_contentfilteredtopic(
   rmw_context_impl_t * const ctx,
   DDS_DomainParticipant * const dp,
   DDS_Topic * const base_topic,
   const char * const cft_name,
   const char * const cft_filter,
+  const rcutils_string_array_t * cft_expression_parameters,
```
asorbini, line 243:
```suggestion
  const rcutils_string_array_t * const cft_expression_parameters,
```

rmw_connextdds_common/src/ndds/dds_api_ndds.cpp
```diff
@@ -1224,3 +1261,120 @@ rmw_connextdds_enable_security(
 
   return RMW_RET_OK;
 }
+
+rmw_ret_t
+rmw_connextdds_set_cft_filter_expression(
+  DDS_TopicDescription * const topic_desc,
+  const char * filter_expression,
```
asorbini, line 1268:
```suggestion
  const char * const filter_expression,
```

rmw_connextdds_common/src/ndds/dds_api_ndds.cpp
```diff
@@ -1224,3 +1261,120 @@ rmw_connextdds_enable_security(
...
+  const rcutils_string_array_t * expression_parameters)
```
asorbini, line 1269:
```suggestion
  const rcutils_string_array_t * const expression_parameters)
```

rmw_connextdds_common/src/rtime/dds_api_rtime.cpp
```diff
@@ -867,13 +867,15 @@ rmw_connextdds_create_contentfilteredtopic(
   DDS_Topic * const base_topic,
   const char * const cft_name,
   const char * const cft_filter,
+  const rcutils_string_array_t * cft_expression_parameters,
```
asorbini, line 870:
```suggestion
  const rcutils_string_array_t * const cft_expression_parameters,
```

rmw_connextdds_common/src/rtime/dds_api_rtime.cpp
```diff
@@ -1952,3 +1954,27 @@ rmw_connextdds_enable_security(
   return RMW_RET_ERROR;
 #endif /* RMW_CONNEXT_ENABLE_SECURITY */
 }
+
+rmw_ret_t
+rmw_connextdds_set_cft_filter_expression(
+  DDS_TopicDescription * const topic_desc,
+  const char * filter_expression,
```
asorbini, line 1961:
```suggestion
  const char * const filter_expression,
```

rmw_connextdds_common/src/rtime/dds_api_rtime.cpp
```diff
@@ -1952,3 +1954,27 @@ rmw_connextdds_enable_security(
...
+  const rcutils_string_array_t * expression_parameters)
```
asorbini, line 1962:
```suggestion
  const rcutils_string_array_t * const expression_parameters)
```

rmw_connextdds_common/src/rtime/dds_api_rtime.cpp
```diff
@@ -1952,3 +1954,27 @@ rmw_connextdds_enable_security(
...
+  UNUSED_ARG(filter_expression);
+  UNUSED_ARG(expression_parameters);
+  return RMW_RET_UNSUPPORTED;
+}
+
+rmw_ret_t
+rmw_connextdds_get_cft_filter_expression(
+  DDS_TopicDescription * const topic_desc,
+  char ** const expr_out,
+  rcutils_string_array_t * cft_params_out);
```
asorbini, line 1974:
```suggestion
  rcutils_string_array_t * const cft_params_out);
```

asorbini (changes requested):
Ok, one last set of changes which I think I will try to commit myself using the "Commit suggestion" features.

They are meant to declare all function arguments as `const`, wherever possible (the only functions which don't do this are the `rmw_api_connextdds_*` ones which have the same signatures as the RMW API).

After this review, I really need to write a "coding style" document for the repository :)

asorbini (approved):
All good 🚀 

I've been thinking `is_cft_supported` might be better renamed to `is_cft_enabled`, but I won't block the PR for that (since that is also related to the API changes in `rmw`).

asorbini:
@iuhilnehc-ynos I have created branch [asorbini/content-filtered-topics](https://github.com/ros2/rmw_connextdds/tree/asorbini/content-filtered-topics) by rebasing your commits on top of `master`, and adding 92bad59.

This commit updates the synchronization of the set/get CFT parameters functions to make sure that they don't overlap with any concurrent WaitSet::wait() (e.g. performed by an executor to which the subscription is attached).

If you want, you can rebase your `topic-content_filtered_topic-rebased` branch on top of this new branch and we can then run some CI to validate this feature and finally merge it.

Btw, you will also need to rebase ros2/rmw#302 in order to build the new branch. I have done so in branch `asorbini/content-filtered-topic` of [asorbini/rmw](https://github.com/asorbini/rmw/tree/asorbini/content-filtered-topics)

iuhilnehc-ynos:
@asorbini 

Thanks for your response.
I'll provide ros2 repos files to run a demo for CFT, please wait a moment.

asorbini:
@iuhilnehc-ynos thank you! I was able to get everything built and I'm about to dig into it, I should have something later today. FYI I had to comment out `rcl/test_subscription_content_filtered_topic_options.cpp` from `rcl/CMakeLists.txt` since the file is not present (it is present in the `rmw` package coincidentally).

asorbini:
@iuhilnehc-ynos Before I could start debugging the concurrency issue, I had a talk with @GerardoPardo and we identified an interesting alternative approach which would eliminate the problem altogether. So I went ahead, implemented it, and created #68 with the resulting changes.

I tested them with the `cft-demo` application (from https://github.com/iuhilnehc-ynos/ros2/blob/topic-debug-connextdds-cft/cft-demo.repos) and using a ROS 2 tree built from the `ros2.repos` you provided (only swapped `rmw_connextdds` for  branch [asorbini/cft] (https://github.com/ros2/rmw_connextdds/tree/asorbini/cft) on the main repository) and everything seemed to work as expected.

If you're ok with it, let's see if this approach can simplify our lives and continue the conversation on that PR.

iuhilnehc-ynos:
Use https://github.com/ros2/rmw_connextdds/pull/68 instead of this PR, close it.

PR #14, opened 2021-03-26
Support extended signature for `message_type_support_callbacks_t::max_serialized_size()` from `rosidl_typesupport_fastrtps_cpp`

MiguelCompany:
This PR adapts the code to the API break introduced by ros2/rosidl_typesupport_fastrtps#67

### DIFF
--- a/rmw_connextdds_common/src/common/rmw_type_support.cpp
+++ b/rmw_connextdds_common/src/common/rmw_type_support.cpp
@@ -677,10 +677,17 @@ void RMW_Connext_MessageTypeSupport::type_info(
   /* The fastrtps type support sets full_bounded to false if unbounded,
      but assumes full_bounded == true by default */
   bool full_bounded = true;
-
+  
+#ifdef ROSIDL_TYPESUPPORT_FASTRTPS_HAS_PLAIN_TYPES
+  char bounds_info;
+  serialized_size_max =
+    static_cast<uint32_t>(callbacks->max_serialized_size(bounds_info));
+  full_bounded = 0 != (bounds_info & ROSIDL_TYPESUPPORT_FASTRTPS_BOUNDED_TYPE);
+#else
   serialized_size_max =
     static_cast<uint32_t>(callbacks->max_serialized_size(full_bounded));
-
+#endif
+  
   unbounded = !full_bounded;
 
   if (full_bounded && serialized_size_max == 0) {
--- a/rmw_connextdds_common/src/common/rmw_type_support.cpp
+++ b/rmw_connextdds_common/src/common/rmw_type_support.cpp
@@ -677,7 +677,7 @@ void RMW_Connext_MessageTypeSupport::type_info(
   /* The fastrtps type support sets full_bounded to false if unbounded,
      but assumes full_bounded == true by default */
   bool full_bounded = true;
-  
+
 #ifdef ROSIDL_TYPESUPPORT_FASTRTPS_HAS_PLAIN_TYPES
   char bounds_info;
   serialized_size_max =
@@ -687,7 +687,7 @@ void RMW_Connext_MessageTypeSupport::type_info(
   serialized_size_max =
     static_cast<uint32_t>(callbacks->max_serialized_size(full_bounded));
 #endif
-  
+
   unbounded = !full_bounded;
 
   if (full_bounded && serialized_size_max == 0) {


asorbini:
Hi @MiguelCompany, thank you for your submission. The changes is small and there are not problems in merging it (as long as we time it with `ros2/rosidl_typesupport_fastrtps#67`).

I have only one request, and I'll [quote myself from another PR](https://github.com/ros2/rmw_connextdds/pull/12#discussion_r601971085).

Basically, at the moment a single branch (`master`) is supporting multiple ROS2 releases. It was the easiest way to handle a quick evolving code base. I plan on creating maintenance branches now that the RMW is more stable, but until then I'd like to keep the branch buildable with multiple versions.

For now, can you add a new "feature flag" in `static_config.hpp`, something like:

```cpp
#ifndef RMW_CONNEXT_HAVE_EXTENDED_TYPE_SUPPORT
#define RMW_CONNEXT_HAVE_EXTENDED_TYPE_SUPPORT \
  (RMW_CONNEXT_RELEASE > RMW_CONNEXT_RELEASE_FOXY)
#endif /* RMW_CONNEXT_HAVE_EXTENDED_TYPE_SUPPORT */
```

And wrap the call to `max_serialized_size()` with it?

(the fact that this is a non-negligible inconvenience for a very small change is definitely a point in favor of introducing branches and getting rid of these macros)

asorbini:
I've actually started the work to get rid of feature flags, and we can merge this PR after I've cleaned up `master` to only track Rolling/Galactic. Please disregard my comment above.

MiguelCompany:
@asorbini I do like feature defines, so what I've done is to add one to ros2/rosidl_typesupport_fastrtps#67 and modified this PR to depend on it.

asorbini:
What?! I get rid of them and you introduce them... 😄 

Jokes aside, I'm also a fan of this strategy in general, but it was getting out of hand for the RMW and it doesn't look like it's much of a common practice in ROS 2 code.

Anyway, I'm ok with it if you think it's useful for `rosidl_typesupport_fastrtps` (e.g., I'm guessing, to build it with different versions of fastrtps). I just [left a minor comment on ros2/rosidl_typesupport_fastrtps#67](https://github.com/ros2/rosidl_typesupport_fastrtps/pull/67/files/c83f80a56f50d72dc4a954b60989fc0009efa815..a48066e8a36be605796b0b2e6ad5a2bc197ea21a#r603465907).

MiguelCompany:
@asorbini I do think it can be removed. Please go ahead with the PR

asorbini:
I created [ros2/rosidl_typesupport_fastrtps#68](https://github.com/ros2/rosidl_typesupport_fastrtps/pull/68) to remove the dependency.

Re: this PR, if you think the feature define will stay in, then it is probably ready to merge.

Btw, thank you for catching the bug in detecting empty messages. I will need to run some tests to validate that everything is fine. The fact that all CI tests are green now makes me think that this might be an undertested scenario (or maybe we just got lucky).

asorbini:
@MiguelCompany do you think this PR will be included in Galactic?

@clalancette thoughts?

MiguelCompany:
@asorbini I hope so.

I suppose someone has to approve it first. And then it should be tested along with  ros2/rosidl_typesupport_fastrtps#67 and ros2/rmw_fastrtps#516

clalancette:
Sorry, this one is going to have to wait for H-Turtle.  It is both an API break and a new feature, so doesn't meet the criteria for Galactic anymore.

MiguelCompany:
@clalancette This PR does not have an API break. It just gets ready for a future API break (the one in ros2/rosidl_typesupport_fastrtps#67). It also fixes a bug [here](https://github.com/ros2/rmw_connextdds/pull/14/files#diff-8310735a1b9a2eb25a4d85563c1a1862ca08f0243488ac9867c76a7ae7d9ed01R659), so I think this could be merged.

We could say the same for ros2/rmw_fastrtps#516

asorbini:
@MiguelCompany good point about the bug fix. It seems to be an undertested situation, since there are no unit tests that I know of failing because of it, but if this PR doesn't make it into Galactic, then I would split that change in a separate fix.

I'll leave it up to @clalancette and others to make a decision, but in theory this PR could also be merged even if ros2/rosidl_typesupport_fastrtps#67 is not merged, since it uses a guard.

asorbini:
I created #33 to merge the bug fix separately.

MiguelCompany:
@clalancette I refactored ros2/rosidl_typesupport_fastrtps#67 to make the change binary compatible

@asorbini I removed the bugfix so that #33 can be merged independently and updated the code to support the new interface. Would you mind reviewing those changes?

asorbini (approved):

Issue #21, opened 2021-03-31
Numerous test failures for Nav2 when using rmw_connextdds


ruffsl:
**System Info**

- _OS_
  - Ubuntu 20.04
- *ROS version and installation type*
  - rolling/source
- *RTI Connext DDS version and installation type*
  - `rti-connext-dds-5.3.1/focal,now 5.3.1-nc.x64Linux3gcc5.4.0+2 amd64`
- *RMW version or commit hash*
  - f46c39d7db6797a80a24631f2890ebd143bfe54e

**Bug Description**

Numerous failures in Nav2 when testing with `rmw_connextdds` that don't occur with other RMWs. To list a few:

* An abnormal number of rmw_dds_common warnings are encountered during testing of nav2:

<details>

```
[controller_server-7] [WARN] [1617199619.684671125] [rmw_dds_common]: nanoseconds value too large for 32-bits, saturated at UINT_MAX
[controller_server-7] [WARN] [1617199619.684736997] [rmw_dds_common]: nanoseconds value too large for 32-bits, saturated at UINT_MAX
[controller_server-7] [WARN] [1617199619.684755779] [rmw_dds_common]: nanoseconds value too large for 32-bits, saturated at UINT_MAX
...
```

</details>

* Expected lexeme type not found as handle's typesupport identifier (rosidl_typesupport_cpp) is not supported by this library:

<details>

```
[planner_server-8] >>> [rcutils|error_handling.c:108] rcutils_set_error_state()
[planner_server-8] This error state is being overwritten:
[planner_server-8] 
[planner_server-8]   'Handle's typesupport identifier (rosidl_typesupport_cpp) is not supported by this library, at /home/jenkins-agent/workspace/packaging_linux/ws/src/ros2/rosidl_typesupport/rosidl_typesupport_cpp/src/type_support_dispatch.hpp:111'
[planner_server-8] 
[planner_server-8] with this new error message:
[planner_server-8] 
[planner_server-8]   'Expected lexeme type (19) not found, search ended at index 21, at /home/jenkins-agent/workspace/packaging_linux/ws/src/ros2/rcl/rcl/src/rcl/lexer_lookahead.c:233'
[planner_server-8] 
[planner_server-8] rcutils_reset_error() should be called after error handling to avoid this.
[planner_server-8] <<<
```

</details>

* Cannot delete and wait on the same object

<details>

```
[...]

PR #20, opened 2021-03-31
Add rmw_publisher_wait_for_all_acked support

Barry-Xu-2018:
Related to ros2/rmw#295

PR #22, opened 2021-04-03
Resolve issues identified while investigating #21

asorbini:
This PR includes several improvements identified while investigating issue #21:

1. **Replace the WaitSet and Conditions implementation based on DDS constructs with one based on the C++ `std` library.**
  - The current implementation causes very high CPU usage when faced with tight loops (e.g. a fast-spinning executor).
    The reason for this is still unclear, but it probably lies in the fundamental mismatch between the ROS 2 WaitSet API and the DDS one. The RMW tries to compensate for the lack of `attach()` and `detach()` in ROS 2 by trying to cache its arguments across `wait()` calls. Evidently this is not enough, and it may be even partially responsible for unnecessarily added CPU cycles under certain circumstances.
  - Instead of trying to fix this implementation further (it also causes somewhat of a "concurrency nightmare" trying to make sure cached objects are not referenced after deletion), the alternative implementation based on C++ `std` was updated to fix the standing issues with "status reporting", and it is now the default.
  - The C++ `std` implementation relies on DDS listeners and coordinates events using `std::condition_variable`'s and `mutex`'s. Because it uses listeners (which automatically consume status flags and reset "change" counters), it must cache and keep track of reported statuses, in order to then later report the correct "status changes" to the user.
2.  **Improved implementation of `RMW_Connext_Client::is_service_available()`**
  - The implementation of `RMW_Connext_Client::is_service_available()` was updated to only return `true` if both a "request reader" and a "reply writer" have been matched from the same remote DomainParticipant.
3. **Support special value `RMW_DURATION_INFINITE` in `WaitSet::wait()`**
  - The RMW earlier assumed that "infinite durations" were only expressed by a null `wait_timeout` argument. I noticed in #20 that there is this constant so I added support for it.
4. **Propagate correct type code for request/reply topics**
  - The code was incorrectly adding a "request header" to the type propagated via discovery for request/reply topics even when the "Extended" mapping was used (the one used by default). This is now fixed.
5. ~**Allow deletion of "waited on" objects (alternative WaitSet implementation)**~
  - ~This implementation will most likely be removed altogether, but I made some updates to try to resolve some "catastrophic failures" when running with applications that seem to be deleting objects while also trying to delete them (`gzserver` from Gazebo). This code didn't fully resolve the issues and is probably still bugged, hence why I ultimately went with item `1.`~
  - ~Note to reviewers: feel free to ignore changes in `rmw_impl_waitset_dds.cpp` and `rmw_waitset_dds.hpp`~
  - The alternative WaitSet implementation, based on DDS objects, was completely removed and saved in branch [asorbini/dds-waitsets](https://github.com/ros2/rmw_connextdds/tree/asorbini/dds-waitsets).

Initial CI validation:
- Linux: [Build Status](https://ci.ros2.org/job/ci_linux/14223/)
- macOS: [Build Status](https://ci.ros2.org/job/ci_osx/11914/)
- Windows: [Build Status](https://ci.ros2.org/job/ci_windows/14327/)


asorbini:
Linter errors resolved in [0b8fd8a](https://github.com/ros2/rmw_connextdds/pull/22/commits/0b8fd8aa076fcaa8cb887fc8ab9231e06b7c669d)

asorbini:
@clalancette @ivanpauno @wjwwood a quick review to merge these changes to `master`, run them through the `nav2` CI again, and still have some time to review the outcome before code freeze, would be greatly appreciated!

CI looks good, and the few errors are linter ones that have already been resolved.

Thanks!

ivanpauno:
>  The current implementation causes very high CPU usage when faced with tight loops (e.g. a fast-spinning executor).
The reason for this is still unclear, but it probably lies in the fundamental mismatch between the ROS 2 WaitSet API and the DDS one. The RMW tries to compensate for the lack of attach() and detach() in ROS 2 by trying to cache its arguments across wait() calls. Evidently this is not enough, and it may be even partially responsible for unnecessarily added CPU cycles under certain circumstances.

This is true, I hope at some point we improve the rmw waitset API.

> This implementation will most likely be removed altogether, but I made some updates to try to resolve some "catastrophic failures" when running with applications that seem to be deleting objects while also trying to delete them (gzserver from Gazebo). This code didn't fully resolve the issues and is probably still bugged, hence why I ultimately went with item 1.

If we're deleting "waited on" objects somewhere, that's a bug.
The rmw implementation doesn't need to support that case.

rmw_connextdds_common/include/rmw_connextdds/rmw_waitset_std.hpp
```diff
@@ -441,14 +456,22 @@ class RMW_Connext_PublisherStatusCondition : public RMW_Connext_StatusCondition
   void update_status_qos(
     const DDS_OfferedIncompatibleQosStatus * const status);
 
-  bool triggered_deadline;
-  bool triggered_liveliness;
-  bool triggered_qos;
+  std::atomic_bool triggered_deadline;
+  std::atomic_bool triggered_liveliness;
+  std::atomic_bool triggered_qos;
```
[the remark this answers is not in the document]
ivanpauno: [triggered_data](https://github.com/ros2/rmw_connextdds/blob/0b8fd8aa076fcaa8cb887fc8ab9231e06b7c669d/rmw_connextdds_common/include/rmw_connextdds/rmw_waitset_std.hpp#L744) seems to be modified but never read (?)

That also looks fishy.

rmw_connextdds_common/src/common/rmw_impl_waitset_dds.cpp
```diff
@@ -721,50 +702,35 @@ RMW_Connext_WaitSet::invalidate(RMW_Connext_Condition * const condition)
...
+  // some packages might not be making the same assumption, so for now, we wait
+  // for the waitset to be free. This might block a thread indefinitely
+  // (if the parallel wait has infinite timeout and never returns).
+  while (this->state != RMW_CONNEXT_WAITSET_FREE) {
+    RMW_CONNEXT_LOG_DEBUG_A(
+      "waiting for waitset to become available for invalidation: "
+      "waitset=%p, condition=%p",
+      static_cast<void *>(this), static_cast<void *>(condition))
+    this->state_cond.wait(lock);
+  }
```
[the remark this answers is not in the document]
asorbini: The problem here is an object deletion (e.g. a subscription, or a publisher, or a guard condition) in parallel with a `wait()` on a WaitSet that was previously used to wait on the same object. This caused the object to be cached in the WaitSet instance and this caused is trying to make sure that the WaitSet doesn't try to access the object again after it's deleted (and this function returns).

Btw, this is the "concurrency nightmare" I mentioned in my initial post.

The original implementation was stricter and threw an error if it detected that a WaitSet had entered a `wait()` with a deleted object still attached to it. I thought this was only "extremely defensive coding" on my part, because I had your same assumption on the validity of such use case (not supported), but it seems it was being triggered by `nav2` tests (see #21, coming from `gzserver`).

I tried running with this code, but I think `gzserver` was ending up in a crash, hence why I stopped trying to fix it, and just switched to the alternative WaitSet implementation.

rmw_connextdds_common/src/ndds/rmw_typecode.cpp
```diff
@@ -948,13 +948,18 @@ rmw_connextdds_convert_type_members(
   DDS_TypeCode * tc_header = nullptr;
 
   if (type_support->type_requestreply()) {
+    // TODO(asorbini) cache condition to a variable to avoid an "unformattable"
+    // long line on the `else()` (see https://github.com/ament/ament_lint/issues/158)
```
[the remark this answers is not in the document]
asorbini: Good to know, thank you for sharing. I'd like to leave it as is in this case, since there is an issue filed for the underlying problem.

rmw_connextdds_common/include/rmw_connextdds/rmw_waitset_std.hpp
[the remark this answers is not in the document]
asorbini: Great catch! I have updated the code to hopefully resolve this race condition, and I also got rid of the `triggered_data` field. See [876ba51](https://github.com/ros2/rmw_connextdds/pull/22/commits/876ba515573e450dd2a1f0ee2cd16c92ecd63dbf)

rmw_connextdds_common/src/common/rmw_impl_waitset_std.cpp
```diff
@@ -702,6 +700,11 @@ RMW_Connext_SubscriberStatusCondition::update_status_deadline(
 {
   this->status_deadline = *status;
   this->triggered_deadline = true;
+
+  this->status_deadline.total_count_change = this->status_deadline.total_count;
+  if (this->reported_deadline) {
+    this->status_deadline.total_count_change -= this->status_deadline_last.total_count;
+  }
```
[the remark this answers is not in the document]
asorbini: Updated in [e090686](https://github.com/ros2/rmw_connextdds/pull/22/commits/e09068699c1e8958101d83a3aeec72251b9e4907)

asorbini:
Addressed review comments. Running CI again to validate:

- ci_linux: [Build Status](https://ci.ros2.org/job/ci_linux/14246/)
- ci_osx: [Build Status](https://ci.ros2.org/job/ci_osx/11936/)
- ci_windows: [Build Status](https://ci.ros2.org/job/ci_windows/14351/)


> If we're deleting "waited on" objects somewhere, that's a bug.
> The rmw implementation doesn't need to support that case.

As I mentioned in [this comment](https://github.com/ros2/rmw_connextdds/pull/22#discussion_r607239434), based on #21, something "fishy" seemed to be happening in Gazebo's `gzserver`, but I haven't debugged it further. It might be that there are still other issues/race condition lingering in the DDS-based implementation that are only triggered by that application. It would be interested to investigate, but I'd say low priority if the current implementation is satisfactory. We might revisit it in the (near) future, to get to the bottom of it, or just get rid of the alternative implementation altogether.

rmw_connextdds_common/src/common/rmw_impl_waitset_std.cpp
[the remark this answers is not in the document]
ivanpauno: thanks!

asorbini:
@ivanpauno after reading the article that you linked, I think the code still suffers from "lost wakeup". Maybe I'm missing something, but I'm considering refactoring to use semaphores and avoid this problem with the condition variable API altogether.

asorbini:
One more run of CI after removing the "lost wakeup" problem 🤞🏼 

- ci_linux: [Build Status](https://ci.ros2.org/job/ci_linux/14255/)
- ci_osx: [Build Status](https://ci.ros2.org/job/ci_osx/11940/)
- ci_windows: [Build Status](https://ci.ros2.org/job/ci_windows/14356/)

rmw_connextdds_common/src/common/rmw_impl.cpp
```diff
@@ -1506,11 +1506,7 @@ RMW_Connext_Subscriber::loan_messages()
     "[%s] loaned messages: %lu",
     this->type_support->type_name(), this->loan_len)
 
-  if (this->loan_len > 0) {
-    return this->status_condition.set_data_available(true);
-  }
-
-  return RMW_RET_OK;
+  return this->status_condition.set_data_available(this->loan_len > 0);
```
ivanpauno, line 1509: just to confirm, do you want to set the condition to `false` if `loan_len == 0`?

(I haven't checked the code well enough to see if that makes sense or not)

rmw_connextdds_common/src/common/rmw_impl.cpp
```diff
@@ -2931,6 +2966,21 @@ RMW_Connext_SubscriberStatusCondition::get_status(
 {
   rmw_ret_t rc = RMW_RET_ERROR;
 
+  // If condition is attached to a waitset, lock its mutex to prevent
+  // modification of the `triggered_*` flags concurrently with WaitSet::wait().
```
ivanpauno, line 2970: I would rather take the mutex in the `get_*` functions, though I don't mind strongly.
`mutex_internal` must be taken before though, if not you cannot ensure `waitset_mutex` (now `mutex_internal` is being taken in the `get_*` functions).

rmw_connextdds_common/src/common/rmw_impl.cpp
```diff
@@ -2982,6 +3032,21 @@ RMW_Connext_PublisherStatusCondition::get_status(
 {
   rmw_ret_t rc = RMW_RET_ERROR;
 
+  // If condition is attached to a waitset, lock its mutex to prevent
```
ivanpauno, line 3035: same as above

ivanpauno (commented):
I still have some concerns related with concurrency.
I also need to take a closer look to the code again...

ivanpauno:
> @ivanpauno after reading the article that you linked, I think the code still suffers from "lost wakeup". Maybe I'm missing something, but I'm considering refactoring to use semaphores and avoid this problem with the condition variable API altogether.

Yes, this kind of code always gets a bit tricky :grimacing: .
The problem with semaphores is that it's not part of the `std` cpp library (until cpp20), and getting multiplatform code (macOS, linux, windows) working doesn't seem to be worth it.

The trick with condition variables is to always follow a pattern like [this](https://github.com/ros2/rmw_connextdds/pull/22#discussion_r607898446) when modifying a "condition" (i.e. something that can modify the return value of `predicate` in `cv.wait(lock, predicate)`).
Not super easy to enforce though :sweat_smile:.

rmw_connextdds_common/include/rmw_connextdds/rmw_waitset_std.hpp
```diff
@@ -84,15 +81,13 @@ class RMW_Connext_Condition
     std::mutex * const waitset_mutex,
     std::condition_variable * const waitset_condition)
   {
-    std::lock_guard<std::mutex> lock(this->mutex_internal);
```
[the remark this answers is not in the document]
asorbini: In theory, it should still be there. I moved the locking to `RMW_Connext_WaitSet::attach()` to perform both the initial "trigger check" and `Condition::attach()` inside `Condition::mutex_internal` without releasing it in between.

Same for `detach()`.

rmw_connextdds_common/include/rmw_connextdds/rmw_waitset_std.hpp
```diff
@@ -133,6 +128,31 @@ class RMW_Connext_Condition
...
+
+    if (nullptr != waitset_mutex) {
+      // Always take the WaitSet's mutex first to prevent deadlocks with
+      // concurrent WaitSet::wait().
+      std::lock_guard<std::mutex> lock_mutex(*waitset_mutex);
+      std::lock_guard<std::mutex> lock(this->mutex_internal);
+      if (nullptr != this->waitset_condition) {
+        this->waitset_condition->notify_one();
+      }
+    }
```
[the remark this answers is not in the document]
asorbini: I thought the problem was that we must guarantee that `std::condition_variable::notify_*()` is called only while the thread performing `WaitSet::wait()` is inside `std::condition_variable::wait()`, otherwise the notification will be lost. Hence why we must synchronize the call to `notify_one()` with the same mutex used (and released) by `condition_variable::wait()`.

Did I misunderstand the problem?

rmw_connextdds_common/include/rmw_connextdds/rmw_waitset_std.hpp
```diff
@@ -184,22 +199,31 @@ class RMW_Connext_GuardCondition : public RMW_Connext_Condition
...
+    auto scope_exit = rcpputils::make_scope_exit(
+      [waitset_mutex]()
+      {
+        if (nullptr != waitset_mutex) {
+          waitset_mutex->unlock();
+        }
+      });
+    if (nullptr != waitset_mutex) {
+      waitset_mutex->lock();
+    }
```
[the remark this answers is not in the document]
asorbini: I think we must establish an order in which the WaitSet's mutex and a Condition's mutex are acquired, otherwise we risk ending up with a deadlock. So far, I was trying to enforce a locking scheme where the WaitSet's mutex is always acquired first, and then a condition's, because this simplifies the implementation of `WaitSet::wait()` and other private functions. It would be possible to invert the order, but then `WaitSet::on_condition_active()` would need to: unlock the WaitSet, lock a Condition, lock the WaitSet, then manipulate the condition, which I thought would be problematic for the whole "lost condition notifications" problem.

Maybe I'm overcomplicating this in my head, I'll review the code again and think more about it. Clearly the current solution is not correct, since there were several test failures in the latest CI.

rmw_connextdds_common/src/common/rmw_impl.cpp
asorbini: Yes, that is the idea. `loan_len` keeps track of the samples loaned from the underlying DDS DataReader (via `DDS_DataReader_take()`), which the RMW caches and goes through them as they are taken by `rmw_take()`.

The flag on the status condition object is supposed to be true until the loan is consumed, returned, and a new one taken if available. As the flag is true, a `wait()` on the associated subscriber should return immediately.

(that's the goal, I suspect something is still not quite right in the implementation though...)

rmw_connextdds_common/include/rmw_connextdds/rmw_waitset_std.hpp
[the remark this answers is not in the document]
ivanpauno: > I think we must establish an order in which the WaitSet's mutex and a Condition's mutex are acquired, otherwise we risk ending up with a deadlock

Yes, I imagined the order was inverted because of a deadlock, but I couldn't see what was it.

>  So far, I was trying to enforce a locking scheme where the WaitSet's mutex is always acquired first, and then a condition's

The problem with that, is that you need to synchronize with `attach()` and `detach()`, so you actually need to do it in the opposite order (if not, `waitset_mutex` and `condition_mutex` might be invalid).

> t would be possible to invert the order, but then WaitSet::on_condition_active() would need to: unlock the WaitSet, lock a Condition, lock the WaitSet, then manipulate the condition, which I thought would be problematic for the whole "lost condition notifications" problem.

Thanks for explaining.
I'm trying to see what is taking the condition `mutex_internal` in `WaitSet::on_condition_active()`, are those `has_data()`, `has_status()`, `has_triggered()`?
If those are the methods taking `mutex_internal`, they shouldn't.
See [this other comment](https://github.com/ros2/rmw_connextdds/pull/22#discussion_r607861543).

`WaitSet::on_condition_active()` is only called from `cv.wait(mutex, on_condition_active)`, i.e. it's called with the `waitset_mutex` hold.
No need to also take the `mutex_internal`, `waitset_mutex` is used to synchronize access to the conditions flags (`triggered_data`, `trigger_value`, etc) and `mutex_internal` is *only* used to synchronize with attach(), detach() (i.e. it's only prottecting the `waitset_mutex` and `condition_mutex` pointers, not the condition flags).

I hope that's clear, sorry if my comments are a bit confusing.

rmw_connextdds_common/include/rmw_connextdds/rmw_waitset_std.hpp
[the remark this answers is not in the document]
ivanpauno: It's not exactly that.

`cv.notify_one()` will do something if someone already called `cv.wait(mutex)` (i.e. a condition variable doesn't have memory).
To avoid losing notifications (and also, the spurios awakes that also `wait()` suffer), an external condition must be used and that condition must be modified with the cv mutex taken:

```cpp
// code notifying
{
  std::lock_guard<std::mutex> guard{cv_mutex};
  cv_flag = true;
}
cv.notify_one();

// other code resetting cv_flag
{
  std::lock_guard<std::mutex> guard{cv_mutex};
  cv_flag = false;
}

// code waiting on the condition
{
  std::unique_lock<std::mutex> guard{cv_mutex};
  while (!cv_flag) {
    cv.wait(mutex);
 }
 // or alternatively
 // cv.wait(mutex, [&cv_flag] (return cv_flag;));
}
```

doing that, you cannot lose modification.
If you were about to enter a `wait()`, two things can happen:

- the flag is set to `true` before taking the cv_mutex, `cv.wait()` will never be called.
- The flag cannot be set to true in the middle of checking the condition and calling `cv.wait()` becuase the `cv_mutex` is taken. The only other possibility is that the flag is set when `cv_wait()` internally released the mutex.
  You can either call `notify_one()` with the mutex taken or not.
  Releasing it before is an "optimization", to avoid the notified thread to awake and contend with the mutex again.

---

This case is more complex, as we have multiple conditions.
The only thing to be careful is to make sure to always take the `cv_mutex` before modifying each of the conditions.

rmw_connextdds_common/src/common/rmw_impl.cpp
ivanpauno: > Yes, that is the idea. loan_len keeps track of the samples loaned from the underlying DDS DataReader (via DDS_DataReader_take()), which the RMW caches and goes through them as they are taken by rmw_take().

Out of curiosity, why is the rmw implementation caching the samples?

It also seems a bit confusing that the subscriber `has_data()`  method can trigger this, I would only expect the condition to be updated from the "on data listener" and reset from the "take_*" methods.

asorbini: > Out of curiosity, why is the rmw implementation caching the samples?

I found it to be a good way to keep track of new available samples in a reader's cache while (re)implementing event handling to support ROS 2 WaitSets.

Neither Pro nor Micro offer an API to query a reader for whether there are new samples in the cache, which is needed to properly manage the `triggered_data` flag inside the RMW, especially if implementing on top of DDS WaitSets, since the DATA_AVAILABLE status is reset at the first `take()`/`read()`, and since ROS 2 doesn't generally consume all available samples once a subscriber is returned by a `wait()`, it is necessary to keep track of this state in the RMW. One could implement it by querying the reader cache with a `read()`, but I thought at that point, it would be best to just `take()` everything available out of the cache, then use the length of the result (`loan_len`) to keep track of whether there are new samples to return (and have `triggered_data` set accordingly).

This `take()` should be relatively inexpensive since there isn't any "deep copy" nor deserialization occurring, only pointers being copied into what are pre-allocated sequences (eventually, since they are allocated as needed but then cached and reused).

> It also seems a bit confusing that the subscriber has_data() method can trigger this, I would only expect the condition to be updated from the "on data listener" and reset from the "take_*" methods.

Querying and caching the reader in one move seems better than accessing the cache multiple times, once to peek if there are samples (to support `has_data()`), and then to actually `take()` samples (to support the `rmw_take_*` methods).

Since the `take()` should be inexpensive, I think it should be fine, although it may be revised if we see some problems.

The samples are only cached as pointers to their serialized buffers, and they are only deserialized "opportunistically" within the `rmw_take_*` operations (see `RMW_Connext_Subscriber::take_next()`).

asorbini:
Pushed a new commit ([b35d211](https://github.com/ros2/rmw_connextdds/pull/22/commits/b35d2114ef3b072f7ba386d1406eff89756d2288)) which should resolve the remaining synchronization issues.

Here's some CI to prove it:
- ci_linux: [Build Status](https://ci.ros2.org/job/ci_linux/14275/)
- ci_osx: [Build Status](https://ci.ros2.org/job/ci_osx/11959/)
- ci_windows: [Build Status](https://ci.ros2.org/job/ci_windows/14375/)

rmw_connextdds_common/src/common/rmw_impl.cpp
asorbini: Removed in [b35d211](https://github.com/ros2/rmw_connextdds/pull/22/commits/b35d2114ef3b072f7ba386d1406eff89756d2288)

rmw_connextdds_common/src/common/rmw_impl.cpp
asorbini: Removed in [b35d211](https://github.com/ros2/rmw_connextdds/pull/22/commits/b35d2114ef3b072f7ba386d1406eff89756d2288)

rmw_connextdds_common/include/rmw_connextdds/rmw_waitset_std.hpp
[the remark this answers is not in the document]
asorbini: Updated in [b35d211](https://github.com/ros2/rmw_connextdds/pull/22/commits/b35d2114ef3b072f7ba386d1406eff89756d2288)

rmw_connextdds_common/include/rmw_connextdds/rmw_waitset_std.hpp
```diff
@@ -84,15 +81,13 @@ class RMW_Connext_Condition
...
   {
-    std::lock_guard<std::mutex> lock(this->mutex_internal);
     this->waitset_mutex = waitset_mutex;
     this->waitset_condition = waitset_condition;
   }
 
   void
   detach()
   {
-    std::lock_guard<std::mutex> lock(this->mutex_internal);
```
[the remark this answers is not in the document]
asorbini: Updated in [b35d211](https://github.com/ros2/rmw_connextdds/pull/22/commits/b35d2114ef3b072f7ba386d1406eff89756d2288)

rmw_connextdds_common/include/rmw_connextdds/rmw_waitset_std.hpp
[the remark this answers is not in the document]
asorbini: Thank you for clarifying, hopefully [b35d211](https://github.com/ros2/rmw_connextdds/pull/22/commits/b35d2114ef3b072f7ba386d1406eff89756d2288) resolves this.

rmw_connextdds_common/include/rmw_connextdds/rmw_waitset_std.hpp
```diff
@@ -164,15 +184,10 @@ class RMW_Connext_GuardCondition : public RMW_Connext_Condition
...
-  trigger_check()
-  {
-    return this->trigger_value.exchange(false);
-  }
-
   bool
   has_triggered()
   {
+    std::lock_guard<std::mutex> lock(this->mutex_internal);
     return this->trigger_value;
```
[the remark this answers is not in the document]
asorbini: Updated in [b35d211](https://github.com/ros2/rmw_connextdds/pull/22/commits/b35d2114ef3b072f7ba386d1406eff89756d2288)

rmw_connextdds_common/include/rmw_connextdds/rmw_waitset_std.hpp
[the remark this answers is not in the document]
asorbini: Updated in [b35d211](https://github.com/ros2/rmw_connextdds/pull/22/commits/b35d2114ef3b072f7ba386d1406eff89756d2288)

rmw_connextdds_common/include/rmw_connextdds/rmw_waitset_std.hpp
```diff
@@ -534,14 +579,38 @@ class RMW_Connext_SubscriberStatusCondition : public RMW_Connext_StatusCondition
...
+      if (nullptr != waitset_mutex) {
+        waitset_mutex->lock();
+      }
+      std::lock_guard<std::mutex> lock(this->mutex_internal);
+      this->triggered_data = true;
+    }
 
-  void
-  on_data();
+    this->notify_waitset();
```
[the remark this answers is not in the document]
asorbini: I introduced `RMW_Connext_Condition::update_state()` in [b35d211](https://github.com/ros2/rmw_connextdds/pull/22/commits/b35d2114ef3b072f7ba386d1406eff89756d2288). Thank you for the suggestion! I applied the same pattern to a couple of other functions too.

asorbini:
CI was clean (only one failure on macOS for flaky `test_multi_threaded_executor`).

Warnings are fixed in [71df269](https://github.com/ros2/rmw_connextdds/pull/22/commits/71df269fd5bc578275a878aba927e964a4c2a11e).

rmw_connextdds_common/include/rmw_connextdds/rmw_waitset_std.hpp
```diff
@@ -698,18 +640,19 @@ class RMW_Connext_SubscriberStatusCondition : public RMW_Connext_StatusCondition
...
-    status->alive_count_change = this->status_liveliness.alive_count_change;
-    status->not_alive_count = this->status_liveliness.not_alive_count;
-    status->not_alive_count_change = this->status_liveliness.not_alive_count_change;
-
-    this->status_liveliness.alive_count_change = 0;
-    this->status_liveliness.not_alive_count_change = 0;
-    this->status_liveliness_last = this->status_liveliness;
+        this->status_liveliness.alive_count_change = 0;
+        this->status_liveliness.not_alive_count_change = 0;
+        this->status_liveliness_last = this->status_liveliness;
```
ivanpauno, line 654: out of curiosity, can't `alive_count` and `alive_count_change` be copied directly from the DDS internal instead of calculated with `status_liveliness_last` and `status_liveliness`?
I don't quite follow why that's needed, though it's also not wrong.

ivanpauno (approved):
LGTM, thanks for all the iterations!!

(the remaining comments can be ignored)

rmw_connextdds_common/include/rmw_connextdds/rmw_waitset_std.hpp
asorbini: The "change" fields must report the change in the other counters from the last time the status was reported to the application. Since the RMW installs listeners on writers and readers, the middleware considers a listener notification as "having reported the status", so the status changes reported to the listeners and cached by the RMW cannot be propagated directly to the application, but rather the RMW must keep track of the last value seen by the application, and basically re-implement what is already done by the middleware. It would be great if there was a way for the listener to not reset the status implicitly, so that these `get_*()` functions could rely on the standard DDS `get_status()` functions, but alas that is not possible at the moment.

rmw_connextdds_common/src/common/rmw_impl_waitset_std.cpp
```diff
@@ -471,19 +496,22 @@ RMW_Connext_WaitSet::wait(
   rmw_events_t * const evs,
   const rmw_time_t * const wait_timeout)
 {
-  std::unique_lock<std::mutex> lock(this->mutex_internal);
-  if (this->waiting) {
-    RMW_CONNEXT_LOG_ERROR_SET(
-      "multiple concurrent wait()s not supported");
-    return RMW_RET_ERROR;
+  // std::unique_lock<std::mutex> lock(this->mutex_internal);
```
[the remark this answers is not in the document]
asorbini: The reason why `attach()` and `detach()` are not called from within the WS's critical section is that I have inverted the order of locking (Condition first, WS second), so it would introduce a deadlock risk with the code executed by the listeners.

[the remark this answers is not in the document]
ivanpauno: ah yes, thanks for explaining!

rmw_connextdds_common/include/rmw_connextdds/rmw_waitset_std.hpp
ivanpauno: makes sense, thanks for explaining!

asorbini:
@clalancette or @wjwwood do you think you might have time to review this so I can merge it? If not, any idea of when that will be? Same for #25 (which is trivial to review).

ivanpauno:
@asorbini can you fix the compiler warnings https://ci.ros2.org/job/ci_osx/11959/clang/?

asorbini:
> Warnings are fixed in [71df269](https://github.com/ros2/rmw_connextdds/pull/22/commits/71df269fd5bc578275a878aba927e964a4c2a11e).

They should already be fixed.

ivanpauno:
> > Warnings are fixed in [71df269](https://github.com/ros2/rmw_connextdds/pull/22/commits/71df269fd5bc578275a878aba927e964a4c2a11e).
> 
> They should already be fixed.

Thanks! I wanted to double check the CI results and I didn't see [this comment](https://github.com/ros2/rmw_connextdds/pull/22#issuecomment-814574381).

asorbini:
CI with [changes](https://github.com/ros2/rmw_connextdds/pull/22/commits/dc75a84d634cd8cc68152ee66a07d2340e5423b4) to resolve some of the test failures detected when running with `rmw_connextdds` as default implementation (see #25):
- ci_linux: [Build Status](https://ci.ros2.org/job/ci_linux/14312/)
- ci_osx: [Build Status](https://ci.ros2.org/job/ci_osx/11998/)
- ci_windows: [Build Status](https://ci.ros2.org/job/ci_windows/14417/)


Plans are `rmw_connextdds`-only.

asorbini:
I've been investigating the failures:

- `TestAllocatorMemoryStrategy.number_of_entities_with_subscription`: test passes if an exception is added for `rmw_connextdds` [here](https://github.com/ros2/rclcpp/blob/master/rclcpp/test/rclcpp/strategies/test_allocator_memory_strategy.cpp#L514). I will create a PR to fix it.
- `TestInteractiveMarkerClient.states`: this test passes if the default "heartbeat period" is increased to have more responsive reliability protocol (e.g. with an [XML file like the one I posted here](https://github.com/ros2/rmw_connextdds/issues/21#issuecomment-812770979)). Not sure how to address this. We could extend the "large data optimizations" that I [recently introduced](https://github.com/ros2/rmw_connextdds/commit/0546ce9126f69700fb0b03d23562856aac2a558c) (or something similar) to apply to all endpoints, but we should review the impact of such a decision.
- `rclpy/test_messages`: this test has a [bug](https://github.com/ros2/rclpy/blob/master/rclpy/test/test_messages.py#L68), since it creates a writer for type `BasicTypes` and tries to publish a `Strings` message. I can create a PR to fix it.
- `TestGraphAPI__rmw_connextdds`: I accidentally deleted too much in my last commit. Restoring a couple of topic name validations makes the failing tests pass.
- `message_filters` tests: I'm unable to reproduce these failures.

asorbini:
PRs to fix `rclcpp/TestAllocatorMemoryStrategy` and `rclpy/test_messages`:
- [rclcpp#1625](https://github.com/ros2/rclcpp/pull/1625)
- [rclpy#764](https://github.com/ros2/rclpy/pull/764)

`TestGraphAPI` fixed in [7de7492](https://github.com/ros2/rmw_connextdds/pull/22/commits/7de74922956ab48057cdbcf1fb5c33751785bc85)

asorbini:
re: `rclcpp.TestExecutor.add_remove_node_thread_safe`, `valgrind` confirms that a guard condition is accessed after being disposed:

```
Expires on 00-jan-00 See www.rti.com for more information.
==54549== Thread 3:
==54549== Invalid read of size 4
==54549==    at 0x5399FC4: pthread_mutex_lock (pthread_mutex_lock.c:67)
==54549==    by 0x18E2C3: __gthread_mutex_lock(pthread_mutex_t*) (gthr-default.h:749)
==54549==    by 0x18EBF7: std::mutex::lock() (std_mutex.h:100)
==54549==    by 0x191F6B: std::lock_guard<std::mutex>::lock_guard(std::mutex&) (std_mutex.h:159)
==54549==    by 0x5DA9B7D: void RMW_Connext_Condition::detach<RMW_Connext_WaitSet::detach(rmw_subscriptions_t*, rmw_guard_conditions_t*, rmw_services_t*, rmw_clients_t*, rmw_events_t*, unsigned long&)::{lambda()#6}>(RMW_Connext_WaitSet::detach(rmw_subscriptions_t*, rmw_guard_conditions_t*, rmw_services_t*, rmw_clients_t*, rmw_events_t*, unsigned long&)::{lambda()#6}&&) (rmw_waitset_std.hpp:99)
==54549==    by 0x5DA759C: RMW_Connext_WaitSet::detach(rmw_subscriptions_t*, rmw_guard_conditions_t*, rmw_services_t*, rmw_clients_t*, rmw_events_t*, unsigned long&) (rmw_impl_waitset_std.cpp:459)
==54549==    by 0x5DA7B25: RMW_Connext_WaitSet::wait(rmw_subscriptions_t*, rmw_guard_conditions_t*, rmw_services_t*, rmw_clients_t*, rmw_events_t*, rmw_time_t const*) (rmw_impl_waitset_std.cpp:543)
==54549==    by 0x5DBC0E0: rmw_api_connextdds_wait(rmw_subscriptions_t*, rmw_guard_conditions_t*, rmw_services_t*, rmw_clients_t*, rmw_events_t*, rmw_wait_set_t*, rmw_time_t const*) (rmw_waitset.cpp:142)
==54549==    by 0x51577F8: rmw_wait (rmw_api_impl_ndds.cpp:800)
==54549==    by 0x55DF853: rmw_wait (functions.cpp:491)
==54549==    by 0x511788F: rcl_wait (wait.c:609)
==54549==    by 0x4E2A56D: rclcpp::Executor::wait_for_work(std::chrono::duration<long, std::ratio<1l, 1000000000l> >) (executor.cpp:738)
==54549==  Address 0xe970ce8 is 24 bytes inside a block of size 80 free'd
==54549==    at 0x48451CF: operator delete(void*, unsigned long) (in /usr/lib/x86_64-linux-gnu/valgrind/vgpreload_memcheck-amd64-linux.so)
==54549==    by 0x5DA49A2: RMW_Connext_GuardCondition::~RMW_Connext_GuardCondition() (rmw_waitset_std.hpp:185)
==54549==    by 0x5D9E60E: rmw_connextdds_destroy_guard_condition(rmw_guard_condition_t*) (rmw_impl.cpp:1886)
==54549==    by 0x5DBB981: rmw_api_connextdds_destroy_guard_condition(rmw_guard_condition_t*) (rmw_waitset.cpp:62)
==54549==    by 0x5157740: rmw_destroy_guard_condition (rmw_api_impl_ndds.cpp:762)
==54549==    by 0x55DF69E: rmw_destroy_guard_condition (functions.cpp:471)
==54549==    by 0x510035A: rcl_guard_condition_fini (guard_condition.c:132)
==54549==    by 0x4E843FC: rclcpp::node_interfaces::NodeBase::~NodeBase() (node_base.cpp:163)
==54549==    by 0x4E845CD: rclcpp::node_interfaces::NodeBase::~NodeBase() (node_base.cpp:169)
==54549==    by 0x4E7BADF: std::_Sp_counted_ptr<rclcpp::node_interfaces::NodeBase*, (__gnu_cxx::_Lock_policy)2>::_M_dispose() (shared_ptr_base.h:377)
==54549==    by 0x197FA3: std::_Sp_counted_base<(__gnu_cxx::_Lock_policy)2>::_M_release() (shared_ptr_base.h:155)
==54549==    by 0x191D6E: std::__shared_count<(__gnu_cxx::_Lock_policy)2>::~__shared_count() (shared_ptr_base.h:730)
==54549==  Block was alloc'd at
==54549==    at 0x48440F3: operator new(unsigned long, std::nothrow_t const&) (in /usr/lib/x86_64-linux-gnu/valgrind/vgpreload_memcheck-amd64-linux.so)
==54549==    by 0x5D9E289: rmw_connextdds_create_guard_condition(bool) (rmw_impl.cpp:1860)
==54549==    by 0x5DBB7A2: rmw_api_connextdds_create_guard_condition(rmw_context_t*) (rmw_waitset.cpp:40)
==54549==    by 0x5157722: rmw_create_guard_condition (rmw_api_impl_ndds.cpp:754)
==54549==    by 0x55DF641: rmw_create_guard_condition (functions.cpp:466)
==54549==    by 0x51000FE: __rcl_guard_condition_init_from_rmw_impl (guard_condition.c:86)
==54549==    by 0x510021E: rcl_guard_condition_init (guard_condition.c:108)
==54549==    by 0x4E83AE3: rclcpp::node_interfaces::NodeBase::NodeBase(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::shared_ptr<rclcpp::Context>, rcl_node_options_t const&, bool, bool) (node_base.cpp:51)
==54549==    by 0x4E6F670: rclcpp::Node::Node(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, rclcpp::NodeOptions const&) (node.cpp:153)
==54549==    by 0x4E6EED2: rclcpp::Node::Node(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, rclcpp::NodeOptions const&) (node.cpp:100)
==54549==    by 0x1A95CF: void __gnu_cxx::new_allocator<rclcpp::Node>::construct<rclcpp::Node, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > >(rclcpp::Node*, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >&&) (new_allocator.h:147)
==54549==    by 0x1A8F77: void std::allocator_traits<std::allocator<rclcpp::Node> >::construct<rclcpp::Node, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > >(std::allocator<rclcpp::Node>&, rclcpp::Node*, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >&&) (alloc_traits.h:484)
==54549== 

```

rmw_connextdds_common/src/common/rmw_graph.cpp
```diff
@@ -744,7 +746,7 @@ rmw_connextdds_graph_add_entityEA(
 
   if (RMW_RET_OK !=
     rmw_connextdds_readerwriter_qos_to_ros(
-      nullptr /* history */,
+      history,
```
ivanpauno, line 749: What test did require this change?
Assuming that `history` policy will be propagated is not correct, so we should fix that test.

(though I don't mind the change here)

rmw_connextdds_common/src/common/rmw_info.cpp
```diff
@@ -235,19 +235,6 @@ rmw_api_connextdds_count_publishers(
...
-    rmw_validate_full_topic_name(topic_name, &validation_result, nullptr);
-  if (RMW_RET_OK != ret) {
-    return ret;
-  }
-  if (RMW_TOPIC_VALID != validation_result) {
-    const char * reason =
-      rmw_full_topic_name_validation_result_string(validation_result);
-    RMW_CONNEXT_LOG_ERROR_A_SET("invalid topic name: %s", reason)
-    return RMW_RET_INVALID_ARGUMENT;
-  }
```
ivanpauno: why is this change required?

rmw_connextdds_common/src/common/rmw_type_support.cpp
```diff
@@ -538,8 +538,25 @@ RMW_Connext_MessageTypeSupport::get_type_support_fastrtps(
...
+      rcutils_reset_error();
+
+      RMW_CONNEXT_LOG_ERROR_A_SET(
+        "failed to load required fastrtps message type support. \n"
+        "Received these errors:\n"
+        "C: '%s'\n"
+        "CPP: '%s'",
+        prev_error_string.str,
+        error_string.str);
+    }
```
ivanpauno, line 559: Ah now I see what https://github.com/ros2/rosidl_typesupport/pull/102 is about.

rmw_connextdds_common/src/common/rmw_info.cpp
ivanpauno: never mind, I see it was restored here.
Though it wasn't restored in `rmw_api_connextdds_get_subscriptions_info_by_topic()` `rmw_api_connextdds_get_publishers_info_by_topic()`

ivanpauno (approved):
I left a few questions in the last changes, LGTM!

I don't mind to include https://github.com/ros2/rmw_connextdds/pull/22/commits/dc75a84d634cd8cc68152ee66a07d2340e5423b4 here, but if there are more fixes needed to make CI greener with only `rmw_connextdds`, I would prefer if they are done in a new PR to make the work of the second reviewer simpler.

rmw_connextdds_common/src/common/rmw_info.cpp
asorbini: I had to remove the check from `rmw_api_connextdds_get_*_info_by_topic()` because otherwise test `TestNodeGraph.get_info_by_topic` would fail:

```
C++ exception with description "Failed to get information by topic for publishers:invalid topic name: topic name must be absolute, it must lead with a '/', at /home/jenkins-agent/workspace/ci_linux/ws/src/ros2/rmw_connextdds/rmw_connextdds_common/src/common/rmw_info.cpp:550, at /home/jenkins-agent/workspace/ci_linux/ws/src/ros2/rcl/rcl/src/rcl/graph.c:677: error not set" thrown in the test body.
```

(from [your CI run for #25](https://ci.ros2.org/job/ci_linux/14302/testReport/rclcpp/TestNodeGraph/get_info_by_topic/))

asorbini: Also, I think it was the same test that required the propagation of `history`, but I have to double check.

asorbini:
CI with new PRs included:

- ci_linux: [Build Status](https://ci.ros2.org/job/ci_linux/14324/)
- ci_osx: [Build Status](https://ci.ros2.org/job/ci_osx/12010/)
- ci_windows: [Build Status](https://ci.ros2.org/job/ci_windows/14428/)

rmw_connextdds_common/src/common/rmw_info.cpp
ivanpauno: > I had to remove the check from rmw_api_connextdds_get_*_info_by_topic() because otherwise test TestNodeGraph.get_info_by_topic would fail:

The validation should only be applied when `no_mangle` is `false`, and skipped when `true`.
I'm almost sure that's why the test was failing, though I haven't tested.

I don't mind if the validation gets completely removed though.

asorbini: I can restore the logic for `!no_mangle`. I'd rather catch it now with these changeset than leaving it for later.

ivanpauno: Sounds good to me

rmw_connextdds_common/src/common/rmw_graph.cpp
asorbini: The test that was failing is [this one](https://github.com/ros2/rclcpp/blob/master/rclcpp/test/rclcpp/node_interfaces/test_node_graph.cpp#L367) which only works because the publisher is local.

ivanpauno: Opened https://github.com/ros2/rclcpp/pull/1626.

asorbini:
> CI with new PRs included:

Failures look in line with my expectations:
- I expected `interactive_markers` tests to fail because of the default reliability protocol settings.
- `sros2.test.sros2.commands.security.verbs.test_generate_policy.test_generate_policy`: this one [failed on osx](https://ci.ros2.org/job/ci_osx/12010/testReport/sros2.test.sros2.commands.security.verbs/test_generate_policy/test_generate_policy/), and I wonder if it's a discovery timing issue.
- `rclcpp.TestExecutor.add_remove_node_thread_safe`: [continued to fail on osx](https://ci.ros2.org/job/ci_osx/12010/testReport/rclcpp/TestExecutor/add_remove_node_thread_safe/), this requires some looking into, especially paired with the [`valgrind` errors](https://github.com/ros2/rmw_connextdds/pull/22#issuecomment-816464838).
- `flake` errors in `rclpy` have already been resolved.
- flaky `rclcpp.TestMultiThreadedExecutor.timer_over_take` flaked again on osx.

ivanpauno:
> flaky rclcpp.TestMultiThreadedExecutor.timer_over_take flaked again on osx.

You can ignore that one, it fails with any rmw implementation.
See https://github.com/ros2/rclcpp/pull/1624.

rmw_connextdds_common/src/common/rmw_info.cpp
asorbini: I don't quite understand the test, but restoring the validation checks only for `!no_mangle` makes [this test](https://github.com/ros2/rclcpp/blob/master/rclcpp/test/rclcpp/node_interfaces/test_node_graph.cpp#L446) fail:

```
[ RUN      ] TestNodeGraph.get_info_by_topic_rcl_remap_topic_name_nullptr
RTI Data Distribution Service Non-commercial license is for academic, research, evaluation and personal use only. USE FOR COMMERCIAL PURPOSES IS PROHIBITED. See RTI_LICENSE.TXT for terms. Download free tools at rti.com/ncl. License issued to Non-Commercial User license@rti.com For non-production use only.
Expires on 00-jan-00 See www.rti.com for more information.
[ERROR] [1618004974.610929217] [rmw_connextdds]: invalid topic name: topic name must not be empty
/home/asorbini/workspace-rmw/issues/21/ws_ros-src/src/ros2/rclcpp/rclcpp/test/rclcpp/node_interfaces/test_node_graph.cpp:446: Failure
Expected: node_graph()->get_publishers_info_by_topic("topic", false) doesn't throw an exception.
  Actual: it throws.
[  FAILED  ] TestNodeGraph.get_info_by_topic_rcl_remap_topic_name_nullptr (329 ms)
```

The complaint from `rmw_connextdds` seems legitimate to me, but somehow the test doesn't expect an exception.

rmw_connextdds_common/src/common/rmw_impl_waitset_dds.cpp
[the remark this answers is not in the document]
hidmic: If we've switched to another implementation, why commit this (partial changes)?

hidmic (commented):
Overall looks reasonable to me too.

rmw_connextdds_common/src/common/rmw_info.cpp
```diff
@@ -538,19 +538,6 @@ rmw_api_connextdds_get_publishers_info_by_topic(
...
```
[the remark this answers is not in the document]
asorbini: See [this thread](https://github.com/ros2/rmw_connextdds/pull/22#discussion_r610673301) about these removals. I'm considering restoring these checks, only for `!no_mangle`, but a unit test fails (which is actually the reason why I had removed them in the first place, because that test was failing, and I assumed the test to be correct since I noticed that other RMW implementation don't perform the same validation).

rmw_connextdds_common/src/common/rmw_impl_waitset_dds.cpp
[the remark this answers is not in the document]
asorbini: The reason for the partial changes in this PR is that when I started investigating #21, I first tried to fix the DDS-based implementation, before realizing that it was causing too much CPU usage anyway, and that fixing the C++ std one would have probably ended up getting us a "better return for the investment".

I could have just deleted the DDS-based implementation, but at this point I've switched between the two implementations back and forth several times, and I didn't want to delete it again, just to have to exhume sometime later in the future.

At the same time though, I doubt we will be reconsidering the implementation in the near future, at least not until either:
- I find time to investigate the performance bottlenecks in more detail, and possibly identify improvements in the Connext libraries.
- ROS 2 introduces `attach()/detach()` and we can re-evaluate the code without all the "hackish" optimizations to try to cope with the lack of these functions in the API.

So maybe, the best thing would be to preserve this implementation in a (dead for now) branch, and clean up the code to only include the one that is actually used.

I can modify the PR to this end, and create a personal branch to preserve the DDS-based WaitSet implementation.

asorbini:
Removed DDS-based implementation altogether in [ce3f86d](https://github.com/ros2/rmw_connextdds/pull/22/commits/ce3f86d525076d1f6434d8cabfd78921494ce42f), and created branch [asorbini/dds-waitsets](https://github.com/ros2/rmw_connextdds/tree/asorbini/dds-waitsets) to preserve it for future reference.

asorbini:
I opened [rmw_connextdds#26](https://github.com/ros2/rmw_connextdds/pull/26) to introduce some QoS optimizations for the reliability protocol which resolve the test failures in `interactive_markers`.

asorbini:
I've been investigating the [invalid memory access in `rclcpp.TestExecutor.add_remove_node_thread_safe`](https://github.com/ros2/rmw_connextdds/pull/22#issuecomment-816464838) and through some `printf()`'s and `valgrind` I was able to confirm that a guard condition is being accessed after having been deleted:

```
******* WAITSET WAIT [BEGIN] 0x5ab20b0 ******
******* GUARD CONDITION [ATTACH] 0x5ab1650 ******
******* GUARD CONDITION [ATTACH] 0x5ab17c0 ******
******* GUARD CONDITION [ATTACH] 0xea81220 ******
###### EXECUTOR NODE [REMOVE] node_1
******* GUARD CONDITION [DELETE] 0xedaadd0 ******
******* GUARD CONDITION [DELETE] 0xeda6960 ******
###### NODE BASE GUARD CONDITION [FINI] 0xea810c0
******* GUARD CONDITION [DELETE] 0xea81220 ******
###### EXECUTOR NODE [NEW] node_2
###### NODE BASE GUARD CONDITION [INIT] 0xf38f9b0
******* GUARD CONDITION [NEW] 0xf38fa90 ******
******* NODE [NEW] 0xedca270 ******
******* GUARD CONDITION [DETACH] 0x5ab1650 ******
******* GUARD CONDITION [DETACH] 0x5ab17c0 ******
******* GUARD CONDITION [DETACH] 0xea81220 ******
==203993== Thread 3:
==203993== Invalid read of size 4
==203993==    at 0x539AFC4: pthread_mutex_lock (pthread_mutex_lock.c:67)
==203993==    by 0x18E389: __gthread_mutex_lock(pthread_mutex_t*) (gthr-default.h:749)
==203993==    by 0x18ECBD: std::mutex::lock() (std_mutex.h:100)
==203993==    by 0x192031: std::lock_guard<std::mutex>::lock_guard(std::mutex&) (std_mutex.h:159)
==203993==    by 0x5DAD111: void RMW_Connext_Condition::detach<RMW_Connext_WaitSet::detach(rmw_subscriptions_t*, rmw_guard_conditions_t*, rmw_services_t*, rmw_clients_t*, rmw_events_t*, unsigned long&)::{lambda()#6}>(RMW_Connext_WaitSet::detach(rmw_subscriptions_t*, rmw_guard_conditions_t*, rmw_services_t*, rmw_clients_t*, rmw_events_t*, unsigned long&)::{lambda()#6}&&) (rmw_waitset_std.hpp:98)
==203993==    by 0x5DAAAD0: RMW_Connext_WaitSet::detach(rmw_subscriptions_t*, rmw_guard_conditions_t*, rmw_services_t*, rmw_clients_t*, rmw_events_t*, unsigned long&) (rmw_impl_waitset_std.cpp:460)
==203993==    by 0x5DAB0BA: RMW_Connext_WaitSet::wait(rmw_subscriptions_t*, rmw_guard_conditions_t*, rmw_services_t*, rmw_clients_t*, rmw_events_t*, rmw_time_t const*) (rmw_impl_waitset_std.cpp:546)
==203993==    by 0x5DBF93B: rmw_api_connextdds_wait(rmw_subscriptions_t*, rmw_guard_conditions_t*, rmw_services_t*, rmw_clients_t*, rmw_events_t*, rmw_wait_set_t*, rmw_time_t const*) (rmw_waitset.cpp:142)
==203993==    by 0x5158838: rmw_wait (rmw_api_impl_ndds.cpp:800)
==203993==    by 0x55E0853: rmw_wait (functions.cpp:492)
==203993==    by 0x511935C: rcl_wait (wait.c:609)
==203993==    by 0x4E2A58D: rclcpp::Executor::wait_for_work(std::chrono::duration<long, std::ratio<1l, 1000000000l> >) (executor.cpp:738)
```
I think the problem is that when a Node object is removed from the executor and ultimately deleted, there is no guarantee that the executor thread has already got out of a concurrent `wait()` before the Node and its "notify guard condition" are deleted.

When the executor's waitset is awaken, it goes through its list of condition to check their status and "tear down" the `wait()` but it instead ends up accessing invalid memory.

Relevant code in rclcpp:
- [test_executor.cpp](https://github.com/ros2/rclcpp/blob/master/rclcpp/test/rclcpp/test_executor.cpp#L110)
- [static_single_thread_executor.cpp](https://github.com/ros2/rclcpp/blob/master/rclcpp/src/rclcpp/executors/static_single_threaded_executor.cpp#L189)
- [node_base.cpp](https://github.com/ros2/rclcpp/blob/master/rclcpp/src/rclcpp/node_interfaces/node_base.cpp#L163)

Not sure about a fix, because I'm not too familiar with `rclcpp`, but I think that the executor should implement some type of synchronization with its thread(s) to make sure that any concurrent WaitSet::wait() that was associated with a removed node has returned before `StaticSingleThreadedExecutor::remove_node()` returns.

For now, I don't think this is a bug in `rmw_connextdds` but an error in `rclcpp` since it stems from a concurrent deletion and `wait()`.

I wonder if this is the root cause of the concurrent `wait()`'s and deletions that were detected with the previous DDS-based implementation.

ivanpauno:
> For now, I don't think this is a bug in rmw_connextdds but an error in rclcpp since it stems from a concurrent deletion and wait().

Yes, your description sounds like a probelem in `rclcpp`.
I will try to figure out what the problem is.

rmw_connextdds_common/src/common/rmw_info.cpp
asorbini: @ivanpauno @hidmic any input on this issue?

Can you confirm if my assumption that the test should in fact expect an exception because the topic name is remapped to an empty name is correct?

If so, I could reintroduce the checks in `rmw_api_connextdds_get_subscriptions_info_by_topic()` `rmw_api_connextdds_get_publishers_info_by_topic()`, but other RMW implementation will likely need to be updated.

I can also open a PR to fix the test and make reintroduce the checks in `rmw_connextdds` in a separate PR.

If, on the other hand, the test is right, then I guess the code is good "as is".

asorbini:
@ivanpauno thanks, looking forward to your findings!

rmw_connextdds_common/src/common/rmw_info.cpp
hidmic: TL; DR let's leave the checks out.

So, the test itself is odd. Its name suggests it is mocking to force a [no matching rule case](https://github.com/ros2/rcl/blob/414fdbd479b83f2b8af8235ad1fa3aeba2b0027f/rcl/include/rcl/remap.h#L101-L103) but then yields an empty string, which is an invalid remapping (and impossible to construct, for that matter). That gets forwarded to the `rmw` implementation, where the graph query yields nothing. They could be doing these checks, all the information is there, but alas, the `rmw` API contract [does not impose any restriction](https://github.com/ros2/rmw/blob/a6420837ccfb66b9df8addc7c1dadd50f39ec74a/rmw/include/rmw/get_topic_endpoint_info.h#L26-L89) (an omission, most likely). 

For the time being, I think the best move is to stay consistent with the API contract and not do any checks. We can improve the test and extend the API contract for H-Turtle. Sorry for the back and forth @asorbini.

hidmic (approved):

rmw_connextdds_common/src/common/rmw_info.cpp
asorbini: Thank you for the explanation @hidmic, I agree that leaving the checks out is the best solution for now.

ivanpauno:
This is ready to merge, one extra CI run to double check last changes are ok:

- ci_linux: [Build Status](https://ci.ros2.org/job/ci_linux/14353/)
- ci_osx: [Build Status](https://ci.ros2.org/job/ci_osx/12033/)
- ci_windows: [Build Status](https://ci.ros2.org/job/ci_windows/14452/)

ivanpauno:
Wow, this is really reducing the ammount of failures when testing with `rmw_connextdds` only.

Great work @asorbini !!!

PR #41, opened 2021-04-22
Add environment variable to control override of DomainParticipantQos

asorbini:
This PR is similar to #7 but focuses on the DomainParticipant's QoS.

At the moment, `rmw_connextdds` loads the default DomainParticipantQos from the DomainParticipantFactory, and it applies some "hard-coded" optimizations:
- Based on ROS 2 options, apply "localhost only" and "enclave" settings.
- Modify some "internal" policies to optimize the "out of the box" experience (e.g. increase maximum size of type objects shared via discovery, and speed up discovery of endpoint).

While it is already possible to disable some of this optimizations specifically (e.g. using variable `RMW_CONNEXT_DISABLE_FAST_ENDPOINT_DISCOVERY`), there is no way for a user to completely disable them altogether, making it impossible to customize these parameters in XML.

This PR introduces a new variable, `RMW_CONNEXT_PARTICIPANT_QOS_OVERRIDE_POLICY`, to control how the RMW will override the default DomainParticipantQos obtained from the DomainParticipantFactory. The variable supports the following values:
- `all` (default if unspecified): apply all customizations.
- `basic`: only apply those changes derived from ROS 2 options.
- `never`: do not modify the default value at all.

I have also consolidated the code to lookup variables `RMW_CONNEXT_ENDPOINT_QOS_OVERRIDE_POLICY` and `RMW_CONNEXT_INITIAL_PEERS` to `rmw_api_connextdds_init()` next to all other environment variables.


asorbini:
This PR is for post-Galactic of course, but I think it would make sense to consider it for a patch release.

rmw_connextdds_common/src/common/rmw_context.cpp
```diff
@@ -91,35 +91,15 @@ rmw_connextdds_initialize_participant_qos(
...
-      false /* append_values */);
-    if (RMW_RET_OK != rc) {
-      RMW_CONNEXT_LOG_ERROR_A(
-        "failed to parse initial peers: '%s'", initial_peers)
-      return rc;
+  if (ctx->participant_qos_override_policy ==
+    rmw_context_impl_t::participant_qos_override_policy_t::All &&
+    DDS_StringSeq_get_length(&ctx->initial_peers) > 0)
+  {
+    printf("custom peers!\n");
```
ivanpauno, line 98: I guess this is a leftover from debugging

```suggestion
```

rmw_connextdds_common/src/common/rmw_context.cpp
```diff
@@ -922,6 +968,37 @@ rmw_api_connextdds_init(
   ctx->optimize_large_data = '\0' == disable_optimize_large_data_env[0];
 #endif /* RMW_CONNEXT_DEFAULT_LARGE_DATA_OPTIMIZATIONS */
 
+  /* Lookup and configure initial peer from environment */
+  const char * initial_peers = nullptr;
+  lookup_rc =
+    rcutils_get_env(RMW_CONNEXT_ENV_INITIAL_PEERS, &initial_peers);
```
ivanpauno, line 974: couldn't `NDDS_DISCOVERY_PEERS` be used directly?

ivanpauno:
> This PR is for post-Galactic of course

There's a `galactic` branch now, so there's no more freeze.

> but I think it would make sense to consider it for a patch release.

SGTM

rmw_connextdds_common/src/common/rmw_context.cpp
asorbini: Oops, fixed in 4e4fd65

rmw_connextdds_common/src/common/rmw_context.cpp
asorbini: Yes, it can be, and it probably should be for `rmw_connextdds`. `RMW_CONNEXT_INITIAL_PEERS` was introduced mostly to support `rmw_connextddsmicro` which doesn't have the capability of configuring initial peers from environment itself.

Btw, this is [mentioned in the README section](https://github.com/ros2/rmw_connextdds/blob/d3c3265a92f6294e67376f837ecf8d37f0de7c24/README.md#RMW_CONNEXT_INITIAL_PEERS), inviting Pro users to use `NDDS_DISCOVERY_PEERS`:

> While both `rmw_connextdds` and `rmw_connextddsmicro` will honor this variable,
[equivalent, and more advanced, functionality is already available in RTI Connext DDS](https://community.rti.com/static/documentation/connext-dds/6.0.1/doc/manuals/connext_dds/html_files/RTI_ConnextDDS_CoreLibraries_UsersManual/Content/UsersManual/ConfigPeersListUsed_inDiscov.htm),
for example using variable `NDDS_DISCOVERY_PEERS`.
>
> For this reason, only users of `rmw_connextddsmicro` should consider specifying
`RMW_CONNEXT_INITIAL_PEERS`.

ivanpauno: > Btw, this is mentioned in the README section, inviting Pro users to use NDDS_DISCOVERY_PEERS:

Ah sorry, I didn't see that when reviewing.

ivanpauno (approved):

rmw_connextdds_common/src/common/rmw_context.cpp
asorbini: No problem, just wanted to point it out for completeness!

asorbini:
CI validation (`rmw_connextdds` only):
- Linux [Build Status](https://ci.ros2.org/job/ci_linux/14483/)
- macOS [Build Status](https://ci.ros2.org/job/ci_osx/12150/)
- Windows [Build Status](https://ci.ros2.org/job/ci_windows/14584/)

asorbini:
Merging since all failures are unrelated.

Only unexpected one is [`TestService__rmw_connextdds.send_reponse_with_client_gone` on macOS](https://ci.ros2.org/job/ci_osx/12150/testReport/(root)/projectroot/test_service__rmw_connextdds/), but that seems more related to timing.

PR #49, opened 2021-04-27
Backport #41 to branch `dashing`

asorbini:
This PR backports changes from #41 to branch `dashing`.


asorbini:
I have verified locally that the code builds and that things work with a hello world. I will merge the PR without additional reviews, since Dashing is not supported as Tier 1.

PR #51, opened 2021-04-27
update guard condition by managing multiple items

iuhilnehc-ynos:
related to https://github.com/ros2/rclcpp/issues/1611


fujitatomoya:
@asorbini friendly ping, when you got time.

asorbini:
@iuhilnehc-ynos @fujitatomoya I haven't read through that whole thread, but can you summarize why a guard condition should be allowed to be attached to multiple WaitSets?

This seems like a big change, and it's inconsistent with other "waitables" in ROS 2 (that can only be attached to one waitset at a time). This is a fundamental contract being broken and the whole implementation of WaitSets and Conditions will require reviewing.

I'm a bit tired of addressing this code in the RMW, so apologies for the push back.

iuhilnehc-ynos:
not a correct fix, close it

PR #53, opened 2021-04-28
Always modify UserObjectQosPolicy regardless of override policy

asorbini:
This PR updates `rmw_connextdds` to always modify `DomainParticipantQos::user_object` (as introduced by #25) regardless of the participant QoS override policy requested by the user (introduced by #41).

This is required to let the participant be shared with applications using Connext's C++11 API, even when they use policies `basic` or `never`. In these cases, applications would not be able to properly use the participant because UserObjectQosPolicy cannot be modified from XML:
```
RTIXMLParser_validateOnStartTag:Parse error at line 263: Unexpected tag 'user_object'
```

The change is trivial, since it only consists of moving a block to the beginning of `rmw_connextdds_initialize_participant_qos_impl()`.

### DIFF
--- a/rmw_connextdds_common/src/ndds/dds_api_ndds.cpp
+++ b/rmw_connextdds_common/src/ndds/dds_api_ndds.cpp
@@ -88,6 +88,24 @@ rmw_connextdds_initialize_participant_qos_impl(
   rmw_context_impl_t * const ctx,
   DDS_DomainParticipantQos * const dp_qos)
 {
+#if RMW_CONNEXT_SHARE_DDS_ENTITIES_WITH_CPP
+  // UserObjectQosPolicy is an internal, undocumented Connext policy used by the
+  // implementations of different language bindings to control the memory
+  // representations of various objects created by user applications. In this
+  // case, the settings match the requirements of the "modern C++" API, and they
+  // allow the DomainParticipant to be used directly by applications that want
+  // to create new entities in C++11, even though the participant was created
+  // using the C API. If these settings are not specified, an application will
+  // receive a SIGSEGV when trying to create one of these entities.
+  //
+  // The customizations are always applied regardless of the QoS override policy
+  // because UserObjectQosPolicy cannot be modified from XML.
+  //
+  dp_qos->user_object.flow_controller_user_object.size = sizeof(void *);
+  dp_qos->user_object.topic_user_object.size = sizeof(void *);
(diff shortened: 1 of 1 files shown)


ivanpauno (approved):

asorbini:
Since this PR has no functional impact wrt/unit testing when compared to #25 and #41 (because the differences only occur when `RMW_CONNEXT_PARTICIPANT_QOS_OVERRIDE_POLICY` is set to a value other than `all`), I'd like to merge it without running CI to save some time, but if we want the "checkbox marked", then I'll start the plans.

@ivanpauno is this allowed by the process?

ivanpauno:
> Since this PR has no functional impact wrt/unit testing when compared to #25 and #41 (because the differences only occur when RMW_CONNEXT_PARTICIPANT_QOS_OVERRIDE_POLICY is set to a value other than all), I'd like to merge it without running CI to save some time, but if we want the "checkbox marked", then I'll start the plans.
>
> @ivanpauno is this allowed by the process?

I agree that there's no functional impact in this PR.
I typically run at least CI in one platform with `colcon build ... --packages-up-to rmw_connextdds_common` `colcon test ... --packages-select rmw_connextdds_common`, to check that everything still builds/linters still pass (I used `rmw_dds_common` in the commands as an example, I write there the list of packages affected by the PR).

asorbini:
Quick CI validation (Linux, build `--packages-up-to rmw_connextdds`, test `--packages-above rti_connext_dds_cmake_module`): [Build Status](https://ci.ros2.org/job/ci_linux/14501/)

ivanpauno:
> --packages-up-to rmw_connextdds, test --packages-above rti_connext_dds_cmake_module

`test --packages-above rti_connext_dds_cmake_module` failed here because it's trying to run tests in packages that weren't build.
You have to make sure that the packages selected in the `test` command are a subset of the ones selected in the `build` command (which can be a bit confusing sometimes).

Linux [Build Status](https://ci.ros2.org/job/ci_linux/14502/)

asorbini:
Oops, thank you for running with the right arguments, I see the issue now. I tried to be "slick" but I was definitely too quick :)

I think I could have used `--packages-above rti_connext_dds_cmake_module --packages-up-to rmw_connextdds` to just test the packages from ros2/rmw_connextdds.

Anyway, since things are green I'll go ahead and merge this PR.

PR #65, opened 2021-09-08
Update rmw_context_impl_t definition

hidmic:
Connected to https://github.com/ros2/rmw/pull/313.

### DIFF
--- a/rmw_connextdds_common/include/rmw_connextdds/context.hpp
+++ b/rmw_connextdds_common/include/rmw_connextdds/context.hpp
@@ -58,6 +58,7 @@ enum class RMW_Connext_RequestReplyMapping
   Extended
 };
 
+// Definition of struct rmw_context_impl_s as declared in rmw/init.h
 struct rmw_context_impl_s
 {
   rmw_dds_common::Context common;
--- a/rmw_connextdds_common/include/rmw_connextdds/context.hpp
+++ b/rmw_connextdds_common/include/rmw_connextdds/context.hpp
@@ -58,7 +58,7 @@ enum class RMW_Connext_RequestReplyMapping
   Extended
 };
 
-struct rmw_context_impl_t
+struct rmw_context_impl_s
 {
   rmw_dds_common::Context common;
   rmw_context_t * base;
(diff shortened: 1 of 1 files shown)


rmw_connextdds_common/include/rmw_connextdds/context.hpp
```diff
@@ -58,7 +58,7 @@ enum class RMW_Connext_RequestReplyMapping
   Extended
 };
 
-struct rmw_context_impl_t
+struct rmw_context_impl_s
```
clalancette, line 61: Same comment as https://github.com/ros2/rmw_fastrtps/pull/558#discussion_r707317216

hidmic: See https://github.com/ros2/rmw_connextdds/pull/65/commits/c447213e916c7114b0f9977821b1363c8b9b113e.

clalancette (approved):

hidmic:
Full CI (from https://github.com/ros2/rmw/pull/313):

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/15218/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9938/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/12898/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/15400/)

asorbini:
@hidmic thank you for the update and apologies for missing out on this PR. I just tried cloning the RMW and build it with the an updated binary installation of Rolling, and I'm experiencing some build failures.

It looks like my local copy of the `rmw` package still defines `rmw_context_impl_t`. Do you know if ros2/rmw#313 has been included in a new release and, if it hasn't, when will it be released?

asorbini:
I see that the changes in `rmw` should have been released already (since ros/rosdistro#30796 was merged), but even after an update, the package is still at version `4.0.0-1focal.20210810.175338`.

clalancette:
> I see that the changes in `rmw` should have been released already (since [ros/rosdistro#30796](https://github.com/ros/rosdistro/pull/30796) was merged), but even after an update, the package is still at version `4.0.0-1focal.20210810.175338`.

Yeah, that's because we've had some core packages fail to build on the buildfarm.  When that happens, nothing gets pushed out.  We're looking at fixing that now.

PR #68, opened 2021-10-22
Add support for user-specified content filters

asorbini:
This PR is a follow up to #12 which takes a different approach in introducing support for content-filtered topics by trying to simplify the implementation and overcome some issues that arose during development of that PR, particularly with respect to concurrency.

The main problem introduced by #12 is the need to delete and re-create the DDS DataReader if the content-filter expression of the associated ROS2 Subscription is modified. Specifically, if the reader was associated with a content-filtered topic and the expression is empty, the reader must be recreated on the base, unfiltered, topic, and viceversa (if the reader didn't have a filter and must now use one, it will be deleted and recreated on a new content-filtered topic).

Beside introducing a whole lot of concurrency problems, a big issue with this strategy is that it makes "enabling/disabling filtering" a very expensive operation both locally (because the existing reader must be finalized and a new one created, causing potentially a whole lot of memory deallocation/allocation), and remotely/on the network (since the deletion of reader will be announced to other participants in order to unmatch it from any remote writer that was communicating with it, and then the new reader will need to be announced and re-matched again).

This is an unfortunate byproduct of the existing DDS API which differentiates between "regular topics" and "content-fitlered topics" instead of making the "filter" a property of the DataReader: if it were possible to create a content-filtered topic with an empty expression (which would cause it to behave like the base unfiltered topic), it would be possible for `rmw_connextdds` to always create a content-filtered topic for every new subscription, even when no filter expression is specified by the user.

In order to enable this use case, we take advantage of RTI Connext DDS' ability to register custom content-filter implementations and create a new custom content-filter which extends the built-in SQL-like filter included in Connext by adding the ability to handle empty filter expressions.

All DataReaders created by `rmw_connextdds` now use a content-filtered topic, and the ROS2 layer can easily manipulate the filtering expression.

The custom content filter is encapsulated in a new package, `rti_connext_dds_custom_sql_filter`.

The filter supports writer-side filtering but it might introduce some minor overhead in the case of an empty expression. This overhead should be characterized through performance testing to make sure any degradation is within an acceptable range.

The branch for this PR was created off of the current `master` (944da9d) with the addition of the implementation for new RMW APIs and some changes to existing internal functions to create content-filtered topics that were introduced by @iuhilnehc-ynos in #12.


ivanpauno (commented):
I don't have bandwith to do a detailed review here, but once we polish the comments in https://github.com/ros2/rmw_fastrtps/pull/513 and https://github.com/ros2/rmw/pull/302 I think it's fine to go ahead with this one (with @iuhilnehc-ynos or @fujitatomoya approval).

fujitatomoya:
@ivanpauno i will review tomorrow.

@iuhilnehc-ynos can you also do final check the implementation just in case.

CC: @asorbini

fujitatomoya (approved):
it looks good to me. but https://github.com/ros2/rmw/pull/302 is under review, so we might need to do some adjustment.

fujitatomoya:
@asorbini @iuhilnehc-ynos 

https://github.com/ros2/design/pull/282#issuecomment-1065393977 is failing because of `rti_connext_dds_custom_sql_filter`.
could you check on that?

iuhilnehc-ynos:
@asorbini 

Could you help check [these unstable cases](https://ci.ros2.org/job/ci_linux/16315/#showFailuresLink) that happened in rmw_connextdds, please?

asorbini:
> @asorbini
> 
> Could you help check [these unstable cases](https://ci.ros2.org/job/ci_linux/16315/#showFailuresLink) that happened in rmw_connextdds, please?

@iuhilnehc-ynos I'll try to reproduce and investigate and get back to you. Thank you for the notification!

fujitatomoya:
@asorbini sorry to bother you, please could you check https://github.com/ros2/rmw/pull/302#discussion_r828725552? the interface `expression_parameters` has been changed a bit from review.

asorbini:
@fujitatomoya @iuhilnehc-ynos I rebased the branch, updated the interfaces and fixed an error that was occurring while finalizing things upon error.

I also checked the failures in the content-filter tests in rclcpp/test_subscription, and realized that the tests should wait for a little bit (e.g. sleep) when they change the filter, to allow the new filter to be propagated to the writer. Otherwise the writer will filter out the new sample before receiving the new filter from the reader, and the sample will not be received because by default historical samples are not re-filtered and repaired (which would also only happen if TRANSIENT_LOCAL durability was used anyway).

I can submit a PR to fix those two tests in rclcpp if you want.

I would appreciate if you could verify that branch `asorbini/cft` of rmw_connextdds works for you and let me know if you have any more problem preventing the CFT feature from being merged.

asorbini:
I don't know if I messed up the review when I rebased the branch, but I now see that all the commits that were added to the repository since the branch was created now show up in the changes, which is pretty annoying.

Not sure if this will create problems with merging the branch back to master, but so far git/GitHub seem all happy about it...

fujitatomoya:
@asorbini it still fails with `rti_connext_dds_custom_sql_filter `, https://ci.ros2.org/job/ci_linux-aarch64/10981/console

```
20:41:01 CMake Error at CMakeLists.txt:11 (find_package):
20:41:01   By not providing "Findrti_connext_dds_cmake_module.cmake" in
20:41:01   CMAKE_MODULE_PATH this project has asked CMake to find a package
20:41:01   configuration file provided by "rti_connext_dds_cmake_module", but CMake
20:41:01   did not find one.
20:41:01 
20:41:01   Could not find a package configuration file provided by
20:41:01   "rti_connext_dds_cmake_module" with any of the following names:
20:41:01 
20:41:01     rti_connext_dds_cmake_moduleConfig.cmake
20:41:01     rti_connext_dds_cmake_module-config.cmake
20:41:01 
20:41:01   Add the installation prefix of "rti_connext_dds_cmake_module" to
20:41:01   CMAKE_PREFIX_PATH or set "rti_connext_dds_cmake_module_DIR" to a directory
20:41:01   containing one of the above files.  If "rti_connext_dds_cmake_module"
20:41:01   provides a separate development package or SDK, be sure it has been
20:41:01   installed.
```

fujitatomoya:
@iuhilnehc-ynos @ivanpauno @wjwwood do you have any clue about this failure?

iuhilnehc-ynos:
CI in [arm64](https://ci.ros2.org/job/ci_linux-aarch64/10981/) ignore the package `rti_connext_dds_cmake_module` even if CI_USE_CONNEXTDDS is enable in [ci_launcher](https://ci.ros2.org/job/ci_launcher/10015/parameters/).

```
# BEGIN SUBSECTION: ignored packages
11:38:45 Trying to ignore the following packages:
...
11:38:45 - rmw_connext_cpp
11:38:45 - rosidl_typesupport_connext_c
11:38:45 - rosidl_typesupport_connext_cpp
11:38:45 - rmw_connext_dynamic_cpp
11:38:45 - connext_cmake_module
11:38:45 - rmw_connext_shared_cpp
11:38:45 - rti_connext_dds_cmake_module
11:38:45 - rmw_connextdds_common
11:38:45 - rmw_connextdds
11:38:45 - rmw_connextddsmicro
11:38:45 - rmw_fastrtps_dynamic_cpp
```

asorbini:
> @asorbini it still fails with `rti_connext_dds_custom_sql_filter `, https://ci.ros2.org/job/ci_linux-aarch64/10981/console

Like @iuhilnehc-ynos noticed, the problem stems from the fact that on arm64 `rmw_connextdds` is added to a "black list" because we currently don't provide a package with libraries for that architecture (although we should probably reconsider this going forward, since Connext does support arm64 of course). The black listing causes [all packages in ros2/rmw_connextdds to be ignored](https://github.com/ros2/ci/blob/9815af79f8edca8a89e3ad913bb6bc5b7a5fb906/ros2_batch_job/__main__.py#L715).

I had forgot about this behavior when encapsulating the custom SQL filter in a new separate package.

One solution would be to update ros2/ci to also include package `rti_connext_dds_custom_sql_filter" in the ignored packages. I don't think that change would require to recreate the plans on the Jenkins server, but I'm not 100% sure. @clalancette and @nuclearsandwich, any thoughts on how feasible this would be with the code freeze coming up?

Another options, possibly easier (at the "expense" of not encapsulating this code in a separate package), would be to get rid of the separate package and just move the code inside `rmw_connextdds_common`.

For now I'll move forward with this second strategy, but for future reference it would be good to know how costly it would be to update that CI script if needed.

Anyway, once again apologies for not considering this problem when introducing a new package. Lesson learned for the future!

fujitatomoya:
@asorbini i just started CI https://github.com/ros2/rmw/pull/302#issuecomment-1073137065, hope this works!

asorbini:
@fujitatomoya thank you for starting another run of CI, but I would try with the commit I just pushed (d7d211c).

While investigating the failures I realized there was a big problem in my strategy when it came to supporting Windows: I had used an internal Connext function which is not exported by the DLLs and thus not available on Windows.

I now overcame that problem by registering the custom filter as a user filter instead of built-in one (using a public function) and to do so, I had to disable writer-side optimizations on Windows.

Nevertheless, this should resolve the build issues on Windows. Fingers crossed and sorry for asking to restart the test!

fujitatomoya:
@asorbini thanks for checking on this!

fujitatomoya:
CI again with https://github.com/ros2/rmw_connextdds/pull/68 (see https://github.com/ros2/rmw_connextdds/pull/68#issuecomment-1073074721)
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/16374/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/10983/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/16745/)

fujitatomoya:
I started another Windows build since I should have now resolved the build issues: [Build Status](https://ci.ros2.org/job/ci_windows/16746/)

fujitatomoya:
@asorbini i will go ahead to merge the stub version https://github.com/ros2/rmw_connextdds/pull/77 to push the interface for CFT.

fujitatomoya:
@iuhilnehc-ynos if you have time, could you take a look at this?

fujitatomoya:
looking at https://ci.ros2.org/job/ci_windows/16746/, i think we just have cpplint minor error https://ci.ros2.org/job/ci_windows/16746/testReport/junit/(root)/projectroot/cpplint/. all the others are unrelated to this PR.

fujitatomoya:
@asorbini i rebased this branch on master with removing stub functions for CFT, and also addressed cpplint error. i will retry the whole CI for this branch.

fujitatomoya:
CI requires default rmw implementation, https://ci.ros2.org/job/ci_linux-aarch64/10995/console.

retry with rmw_fastdds and rmw_connextdds:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/16388/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/10997/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/16760/)

asorbini:
@fujitatomoya thank you for addressing the linter errors! It looks like the arm64 build failed because no RMW was enabled

ivanpauno:
> CI requires default rmw implementation

We don't support Connext in aarch64, so no need to run that one.
In linux/windows we could cancel one of the two jobs that are running, and only keep one.

fujitatomoya:
@asorbini there are still unstable errors, https://ci.ros2.org/job/ci_linux/16388/

fujitatomoya:
windows also unstable,

- https://ci.ros2.org/job/ci_windows/16760/msbuild/folder.-1571461178/ these warnings for rmw_connextdds.
- https://ci.ros2.org/job/ci_windows/16760/msbuild/folder.-696170516/ has been addressed with https://github.com/ros2/rmw_fastrtps/pull/592

asorbini:
I pushed fixes for the warnings and started a new Windows job to confirm it: [Build Status](https://ci.ros2.org/job/ci_windows/16769/)

asorbini:
> @asorbini there are still unstable errors, https://ci.ros2.org/job/ci_linux/16388/
> 
> sometimes I see the core crash with the following back trace.
> test_init__rmw_connextdds
> 
> and if i do not see the coredump, i meet the bunch of following errors in console.
> 
> ```
> ./build/rcl/test/test_init__rmw_connextdds
> ...
> [ERROR] [1647897965.816402612] [rmw_connextdds]: failed to lookup from environment: var=RMW_CONNEXT_USE_DEFAULT_PUBLISH_MODE, rc=some string error
> DDS_DomainParticipantFactory_get_participants:ERROR: Bad parameter: self
> DS_DomainParticipantFactory_unlockI:!precondition: self == ((void *)0)
> DDS_DomainParticipantFactory_get_participants:!unlock factory
> [ERROR] [1647897965.816418234] [rmw_connextdds]: failed to list existing participants
> [ERROR] [1647897965.816420902] [rmw_connextdds]: failed to finalize participant factory
> [ERROR] [1647897965.816422478] [rmw_connextdds]: failed to finalize RMW context
> ```
> 
> i think this is really hard to debug on our side since it seems related to DDS implementation.

Thank you for pointing this out, I will investigate and try to fix it. I believe it has to do with error handling code in the RMW more than internal Connext code. It seems like the RMW is trying to list the local participants, but the DDS factory has not been initialized yet. It might be as easy as adding a check for `factory != NULL` in that code path.

asorbini:
Testing fix for test errors: [Build Status](https://ci.ros2.org/job/ci_linux/16397/)

nuclearsandwich:
> One solution would be to update ros2/ci to also include package `rti_connext_dds_custom_sql_filter" in the ignored packages. I don't think that change would require to recreate the plans on the Jenkins server, but I'm not 100% sure. @clalancette and @nuclearsandwich, any thoughts on how feasible this would be with the code freeze coming up?

I'm getting to this quite late, and so the code freeze deadline is something you'll want to take up with @clalancette but in terms of general feasibility the logic in https://github.com/ros2/ci/blob/9815af79f8edca8a89e3ad913bb6bc5b7a5fb906/ros2_batch_job/__main__.py#L699-L721 just needs to be updated to include additional packages around rmw_connextdds and that does not require a re-deploy as long as the new package does not add any additional dependencies at the system level. If it does add new dependencies, then that is definitely an rmw freeze problem and at this stage I would say it would be best _not_ to do that for Humble.

> ... we currently don't provide a package with libraries for that architecture (although we should probably reconsider this going forward, since Connext does support arm64 of course). The black listing causes [all packages in ros2/rmw_connextdds to be ignored](https://github.com/ros2/ci/blob/9815af79f8edca8a89e3ad913bb6bc5b7a5fb906/ros2_batch_job/__main__.py#L715).

There are several steps to making this happen and I'd love to complete our RHEL support for Connext as well. When you've got cycles for this @asorbini I'd suggest opening an "extending platform support" or similarly phrased issue on rmw_connextdds. The current way we're distributing Pro artifacts is pretty much at its breaking point so I'd like not to add more (such as for arm64) to it without refactoring it.

fujitatomoya:
@asorbini confirmed that https://github.com/ros2/rmw_connextdds/pull/68#issuecomment-1074496290 is also solved with my local environment. thanks!

asorbini:
@fujitatomoya that's great to hear. I think the PR should be ready for being merged 🚀 

@nuclearsandwich Thank you for replying and confirming the changes. I ended up getting rid of that additional module since it didn't add much other than several (albeit minor) headaches. Probably not right away, but I'll follow up on your suggestion to track changes to platform support in an issue, and I'm happy to discuss existing issues and possible solutions either there on in a separate email thread whenever you want.

fujitatomoya:
@asorbini thanks for the fix, this has been merged!

PR #69, opened 2022-01-07
Fix cpplint errors

jacobperron:
Relates to https://github.com/ament/ament_lint/pull/324


asorbini (approved):

clalancette (approved):
Looks good with green CI.

asorbini:
@jacobperron FYI it looks like this PR introduced some linter errors https://build.ros2.org/job/Rdev__rmw_connextdds__ubuntu_focal_amd64/36/testReport/

I see ament/ament_lint#324 has been merged now, so maybe it was just a fluke because of timing. But it has happened for the last two builds so I thought I'd bring it up to your attention.

clalancette:
This is currently happening because we don't have a successful build of Rolling packages at the moment.  Once we fix up a bug there, and re-release some packages, we should have the newer version of cpplint available and this warning will go away in the Rdev builds.

PR #71, opened 2022-02-04
Update rti-connext-dds dependency to 6.0.1.

nuclearsandwich:
Now that this package is available in the ROS bootstrap repository for
Ubuntu Focal and Jammy we can bump the expected dependency version.

Requires https://github.com/ros2/ci/pull/614 and https://github.com/ros/rosdistro/pull/32045

### DIFF
--- a/rti_connext_dds_cmake_module/package.xml
+++ b/rti_connext_dds_cmake_module/package.xml
@@ -10,7 +10,7 @@
   <buildtool_depend>ament_cmake</buildtool_depend>
   <buildtool_export_depend>ament_cmake</buildtool_export_depend>
 
-  <depend>rti-connext-dds-5.3.1</depend>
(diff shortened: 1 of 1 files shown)


clalancette (approved):
I'll note that this change means that we'll only be building binary packages for `rmw_connextdds` for Connext 6.0.1 now.  Users will still be able to build against 5.3.1 from source if they setup their environment to point to 5.3.1 before building this.  So this is exactly the behavior we want.

This looks good to me.  @asorbini any objections?

nuclearsandwich:
> I'll note that this change means that we'll only be building binary packages for `rmw_connextdds` for Connext 6.0.1 now.

And this only applies to Rolling, and by extension eventually Humble since this change won't be backported to the Foxy and Galactic branches.

asorbini (approved):
Looks good, thank you for these updates @nuclearsandwich!

PR #74, opened 2022-03-03
Add sequence numbers to message info structure

ivanpauno:
Implements https://github.com/ros2/rmw/pull/318.

NOTE: I have to update the PR after https://github.com/ros2/rmw/pull/318/commits/78ddbeac21ad467ff35cc3b5dd52c53e94235459

### DIFF
--- a/rmw_connextdds_common/src/common/rmw_impl.cpp
+++ b/rmw_connextdds_common/src/common/rmw_impl.cpp
@@ -1885,6 +1885,10 @@ rmw_connextdds_message_info_from_dds(
   to->source_timestamp = 0;
   to->received_timestamp = 0;
 #endif /* !RTI_WIN32 */
+  // Currently we cannot use `rmw_connextdds_sn_dds_to_ros`, as that was used to convert to
+  // `rmw_request_id_t.sequence_number`, which is an int64_t and not an uint64_t.
+  // When rmw is updated and all sequence numbers are a `uint64_t`,
+  // rmw_connextdds_sn_dds_to_ros() should be updated and used everywhere.
   to->publication_sequence_number =
     static_cast<uint64_t>((from->publication_sequence_number).high) << 32 |
     static_cast<uint64_t>((from->publication_sequence_number).low);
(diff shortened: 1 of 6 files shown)


rmw_connextdds_common/src/common/rmw_impl.cpp
```diff
@@ -1885,8 +1885,12 @@ rmw_connextdds_message_info_from_dds(
...
   to->received_timestamp = 0;
 #endif /* !RTI_WIN32 */
-  rmw_connextdds_sn_dds_to_ros(from->publication_sequence_number, to->publication_sequence_number);
-  rmw_connextdds_sn_dds_to_ros(from->reception_sequence_number, to->reception_sequence_number);
+  to->publication_sequence_number =
+    static_cast<uint64_t>((from->publication_sequence_number).high) << 32 |
+    static_cast<uint64_t>((from->publication_sequence_number).low);
+  to->reception_sequence_number =
+    static_cast<uint64_t>((from->reception_sequence_number).high) << 32 |
+    static_cast<uint64_t>((from->reception_sequence_number).low);
```
asorbini, line 1893: @ivanpauno would it be ok to update macro `rmw_connextdds_sn_dds_to_ros()` to use `static_cast<>`?

I see that it is also used in another place with an `int64_t` value, so that might be a problem, warning wise, but I think we could just update [`RMW_Connext_RequestReplyMessage::sn`](https://github.com/ros2/rmw_connextdds/blob/0bc1cedcd78d3472a62c72c2a4904fe38a511f5e/rmw_connextdds_common/include/rmw_connextdds/type_support.hpp#L46) to be a `uint64_t`.

What do you think?

ivanpauno: The problem is that we will get a warning when converting to [rmw_request_id_t.sequence_number](https://github.com/ros2/rmw/blob/bceeda0a2a1d1451e1e5243fba09aaa3811f33cd/rmw/include/rmw/types.h#L355) if we switch to `uint64_t`.
So I think we need two different conversions.

Ideally, we would modify `rmw_request_id_t.sequence_number` to be an `uint64_t` as well in the future.

asorbini: Oh, I see. Yes, that would probably be a problem, and changing it to uint64_t might be the right solution. In general it's a bit weird to have sequence numbers of different types.

In this case, we can keep the code as is but I would appreciate if you could add a comment that references these reasons for not using `rmw_connextdds_sn_dds_to_ros()` so we can remember them in the future.

jacobperron (approved):

asorbini (approved):

PR #44, opened 2021-04-26
Add rmw listener apis

irobot-ros:
This PR adds stub implementations for the RMW listener APIs.
See [design](https://github.com/ros2/design/pull/305) and [Discourse post](https://discourse.ros.org/t/ros2-middleware-change-proposal/15863).

FYI @wjwwood 

This PR connects to
- design: https://github.com/ros2/design/pull/305
- rclcpp: https://github.com/ros2/rclcpp/pull/1579
- rcl: https://github.com/ros2/rcl/pull/839
- rmw: https://github.com/ros2/rmw/pull/286
- rmw_implementation: https://github.com/ros2/rmw_implementation/pull/161
- rmw_fastrtps: https://github.com/ros2/rmw_fastrtps/pull/468
- rmw_cyclonedds: https://github.com/ros2/rmw_cyclonedds/pull/256

PR #76, opened 2022-03-16
Add support for listener callbacks

asorbini:
This PR implements the new listener-related functions that were introduced by #44.

Signed-off-by: Andrea Sorbini <asorbini@rti.com>


asorbini:
Linux CI with only Connext enabled for initial validation: [Build Status](https://ci.ros2.org/job/ci_linux/16339/)

asorbini:
> Linux CI with only Connext enabled for initial validation: [Build Status](https://ci.ros2.org/job/ci_linux/16339/)

I investigated the failures, I believe the issues lie either in some test problem or they are unrelated to this PR (and they manifested because I ran CI with just this RMW enabled, which usually doesn't happen).

- [rclcpp.TestSubscription.on_new_message_callback](https://ci.ros2.org/job/ci_linux/16339/testReport/rclcpp/TestSubscription/on_new_message_callback/): in the last case, the test only waits while `c3 == 0` while instead the loop should continue for as long as `c3 < 3`.
- [rclcpp.TestWaitForAllAckedWithParm/TestPublisherWaitForAllAcked.check_wait_for_all_acked_with_QosPolicy/0](https://ci.ros2.org/job/ci_linux/16339/testReport/rclcpp/TestWaitForAllAckedWithParm_TestPublisherWaitForAllAcked/check_wait_for_all_acked_with_QosPolicy_0___88_byte_object__01_00_00_00_00_00_00_00_01_00_00_00_00_00_00_00_01_00_00_00_02_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00___88_byte_object__01_00_00_00_00_00_00_00_01_00_00_00_00_00_00_00_01_00_00_00_02_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00________________________/): unrelated issue. Problem is with the test which only waits for 500ms to receive an ACK. Connext's default settings use a 3s periodic heartbeat, which may cause the ACK not to be received within the timeout. Increasing the wait to 6s makes the test pass consistently.
- [rclpy.rclpy.test.test_client.TestClient.test_service_timestamps](https://ci.ros2.org/job/ci_linux/16339/testReport/rclpy.rclpy.test.test_client/TestClient/test_service_timestamps/): unrelated problem. The issue lies with the test which doesn't take into account the possible return value `(None, None)` which indicates a successful `take_request()` with no request actually taken (i.e. an "empty polling" of the requests)..
- [launch_testing_examples.launch_testing_examples.check_node_launch_test](https://ci.ros2.org/job/ci_linux/16339/testReport/launch_testing_examples.launch_testing_examples/check_node_launch_test/launch_testing_examples_check_node_launch_test/) and [launch_testing_examples.launch_testing_examples.set_param_launch](https://ci.ros2.org/job/ci_linux/16339/testReport/launch_testing_examples.launch_testing_examples/set_param_launch_test/launch_testing_examples_set_param_launch_test/): unrelated issue. It seems like the failures can be avoided by increasing the wait timeout or by configuring Connext to more quickly repair samples on topic "ros_discovery_info".
- Errors in `rviz_rendering` seem to be unrelated to the RMW layer because they occur also with other RMW implementations.

ivanpauno:
> [rclcpp.TestSubscription.on_new_message_callback](https://ci.ros2.org/job/ci_linux/16339/testReport/rclcpp/TestSubscription/on_new_message_callback/): in the last case, the test only waits while c3 == 0 while instead the loop should continue for as long as c3 < 3.

Could we fix that test?
It's directly related to the listener callbacks API, so it would be good to get it fixed to make sure that's the only issue.

ivanpauno (approved):
Code LGTM!

asorbini:
> Could we fix that test? It's directly related to the listener callbacks API, so it would be good to get it fixed to make sure that's the only issue.

I opened ros2/rclcpp#1900 to fix the unit tests.

asorbini:
> > Could we fix that test? It's directly related to the listener callbacks API, so it would be good to get it fixed to make sure that's the only issue.
> 
> I opened [ros2/rclcpp#1900](https://github.com/ros2/rclcpp/pull/1900) to fix the unit tests.

Also opened ros2/rclcpp#1901 to fix the issue with the `wait_for_acknowledgments()` test.

rmw_connextdds_common/src/common/rmw_impl_waitset_std.cpp
```diff
@@ -745,6 +750,30 @@ RMW_Connext_SubscriberStatusCondition::update_status_sample_lost(
...
 }
 
+void
+RMW_Connext_SubscriberStatusCondition::notify_new_data()
+{
+  size_t unread_samples = 0;
+  std::unique_lock<std::mutex> lock_mutex(new_data_event_mutex_);
+  perform_action_and_update_state(
+    [this, &unread_samples]() {
+      const rmw_ret_t rc = this->sub->count_unread_samples(unread_samples);
```
alsora, line 760: I just want to double check here, as I don't really know the underlying connext APIs.
When is a sample marked as read?

If this `notify_new_data` function is invoked whenever the DDS receives a message, I would expect `unread_samples` to always be 1.

alsora:
Looks good to me, I just want to confirm how the `notify_new_data` function works exactly.

rmw_connextdds_common/src/common/rmw_impl_waitset_std.cpp
asorbini: > I just want to double check here, as I don't really know the underlying connext APIs.
> When is a sample marked as read?

In general in DDS, samples have a "sample state" associated with them, which is "NOT_READ" when they are added to a DataReader's cache, and it transitions to "READ" once they have been returned at least once by a call to DataReader::read() or DataReader::take() (in the case of take() the "READ" state cannot be really observed because the sample is removed from the cache, but theoretically the sample transitions...).

Both read() and take() take a "sample state" argument to possibly filter the samples in the cache (in addition to two more states, "Instance" and "View". So the `count_unread_samples()` function takes advantage of this to scan the reader's cache and only ever receive samples that haven't yet been observed by the "application layer" (i.e. the RMW/ROS2).

The RMW will then take()'s all the samples out of the cache (when the upper layers call rmw_take_message()) and it does to without constraints on the "sample state".

> If this notify_new_data function is invoked whenever the DDS receives a message, I would expect unread_samples to always be 1.

There is no guarantee in DDS that the callback will be invoked for every samples. In fact the protocol explicitly allows an implementation to coalesce multiple events in a single notification. This is true for every of the observable events, not just DATA_AVAILABLE.

Short of explicit guarantees of an implementation, it is not correct to assume that a DDS listener will be notified for every received sample. The listener must actually inspect the reader cache to determine how many samples are available.

alsora (approved):

rmw_connextdds_common/src/common/rmw_impl_waitset_std.cpp
asorbini: One thing I'd like to point out about this implementation is that there is always a possibility that some samples that were notified to the callback will not actually be read by the application layer because of QoS settings (mainly KEEP_LAST history). The samples could be counted but then be pushed out of the cache by newer samples before they can be take()n out of it.

asorbini:
@clalancette @ivanpauno can this PR be merged now even with the freeze? It doesn't change API/ABI.

audrow:
> @clalancette @ivanpauno can this PR be merged now even with the freeze? It doesn't change API/ABI.

Hi @asorbini, I don't think we should merge this because it is a feature addition and the freeze is in place. If it doesn't break API/ABI, it should be able to be added in a patch release.

cwecht:
Did anyone ever test this PR with irobot's events executors? https://github.com/irobot-ros/events-executor
If I run their tests with rmw_connextdds enabled and these changes here applied to the rmw_connextdds humble branch, I get these kinds of errors/warnings:

grafik

This PR is needed though, if one is trying to use events executors with connextdds. It would be nice if someone could have a look at this issue.

cwecht:
#109 fixes the issue with the events executors.

alsora:
@asorbini FYI we recently opened the PR for the events-executor in rclcpp and we currently have disabled it from running with connext.
To test this PR you can easily remove the "if connext -> skip" parts of the tests 
https://github.com/ros2/rclcpp/pull/2140

asorbini:
@alsora thank you for the heads up, that's very useful to validate these changes.

Do you have a `.repos` file that I can use to test that PR or is it just those `rclcpp` changes + current rollng?

rmw_connextdds_common/src/common/rmw_impl_waitset_std.cpp
cwecht: I this true for both connext dds variantions, Pro and Micro? I'm just asking because the call to count_unread_samples adds quite some overhead here. I don't remember the exact numbers, but I think it was at least 1µs. It's not that much, but it is a gain, which can add up quite quickly for latency sensitive applications with "sufficiently large" chains of subscriptions.

cwecht:
What's the status of this PR? Given that https://github.com/ros2/rclcpp/pull/2155 is merged now, it would be really greate to have this PR "officially" approved and merged in order to use rclcpp with ConnextDDS to it's full potential out of the box.

clalancette:
> What's the status of this PR? Given that [ros2/rclcpp#2155](https://github.com/ros2/rclcpp/pull/2155) is merged now, it would be really greate to have this PR "officially" approved and merged in order to use rclcpp with ConnextDDS to it's full potential out of the box.

At this point, it needs another approval (there have been changes since the previous ones), and CI run on it.  I can do the latter, but I'd prefer @asorbini or @alsora did another review/approval first.

alsora:
I'm not really familiar with connext, but overall the PR looks sane.

From my perspective I would suggest to create a rclcpp branch where the checks that disable events-executor with connext are removed, and then, if it passes all unit-tests, it should be good to go.

cwecht:
@alsora here is the PR to enable the callback groupd tests in rclcpp again: https://github.com/ros2/rclcpp/pull/2182

cwecht:
The tests reenabled in https://github.com/ros2/rclcpp/pull/2182 are now succeeding. Although they seem to be somewhat flakly, I'd suppose, that this PR should be mergable now?

clalancette:
> Although they seem to be somewhat flakly, I'd suppose, that this PR should be mergable now?

We'd really prefer to not merge in "somewhat flaky" tests.  We are just down to handful of those now, and I'd like not to increase that number.

PR #79, opened 2022-03-31
Exclude missing sample info fields when building rmw_connextddsmicro

asorbini:
This is a follow up to #74 which introduced support for some additional message info fields.

In doing so, we accidentally broke the build for `rmw_connextddsmicro` since Connext DDS Micro does not support all of the required fields.

This PR fixes the issue by guarding access to these fields based on the Connext implementation.


ivanpauno (approved):

rmw_connextdds_common/src/common/rmw_impl.cpp
```diff
@@ -1970,9 +1970,11 @@ rmw_connextdds_message_info_from_dds(
   to->publication_sequence_number =
     static_cast<uint64_t>((from->publication_sequence_number).high) << 32 |
     static_cast<uint64_t>((from->publication_sequence_number).low);
+#if RMW_CONNEXT_DDS_API == RMW_CONNEXT_DDS_API_PRO
   to->reception_sequence_number =
     static_cast<uint64_t>((from->reception_sequence_number).high) << 32 |
     static_cast<uint64_t>((from->reception_sequence_number).low);
+#endif  // RMW_CONNEXT_DDS_API == RMW_CONNEXT_DDS_API_PRO
```
clalancette, line 1977: In the `#else` case here, do we need to set `to->reception_sequence_number` to 0 explicitly?

clalancette (commented):
I have one question inline, once we've discussed that this is good to go.

rmw_connextdds_common/src/common/rmw_impl.cpp
ivanpauno: Good catch, it should be set to [RMW_MESSAGE_INFO_SEQUENCE_NUMBER_UNSUPPORTED](https://github.com/ros2/rmw/blob/01d0db60c5bac242fe6f8a6d49494512f1e20c61/rmw/include/rmw/types.h#L525).

ivanpauno (changes requested):
We also need to update https://github.com/ros2/rmw_connextdds/blob/81f34f0d3d5d77eec0c8ec0477b87c8d9bfa9384/rmw_connextdds_common/src/common/rmw_features.cpp#L23-L33 accordingly.

It seems that the common function will not be of much help, [this](https://github.com/ros2/rmw_connextdds/blob/81f34f0d3d5d77eec0c8ec0477b87c8d9bfa9384/rmw_connextddsmicro/src/rmw_api_impl_rtime.cpp#L967-L971) and [this](https://github.com/ros2/rmw_connextdds/blob/81f34f0d3d5d77eec0c8ec0477b87c8d9bfa9384/rmw_connextdds/src/rmw_api_impl_ndds.cpp#L962-L966) will probably need to be implemented directly.

rmw_connextdds_common/src/common/rmw_impl.cpp
asorbini: Fixed in 0c88722

ivanpauno (approved):
LGTM!
Thanks for the fix @asorbini !

ivanpauno:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/16543/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/16917/)

clalancette:
Rebuild now that the missing semicolons have been added:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/16545/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/16919/)

asorbini:
> Rebuild now that the missing semicolons have been added:

Oops, I totally goofed up while validating the change. Thank you for fixing @clalancette!

clalancette:
Here's a full CI, just to validate that things are good before we merge:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/16548/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/11134/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/16922/)

PR #80, opened 2022-04-06
Update launch_testing_ros output filter prefixes for Connext6

ivanpauno:
Hopefully, fixes https://github.com/ros2/ros2cli/issues/705.

### DIFF
--- a/rmw_connextdds/CMakeLists.txt
+++ b/rmw_connextdds/CMakeLists.txt
@@ -73,4 +73,4 @@ ament_package(
 )
 
 ament_index_register_resource("rmw_output_prefixes"
-  CONTENT "RTI Data Distribution Service\nExpires on")
+  CONTENT "RTI Data Distribution Service\nRTI Connext DDS\nExpires on")


ivanpauno:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/16532/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/11119/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/16906/)

clalancette:
Hm, weird.  I guess what I don't understand about this is that the nightly Linux jobs on https://ci.ros2.org seem to be succeeding OK: https://ci.ros2.org/view/nightly/job/nightly_linux_debug/2259/ .  The failure in ros2/ros2cli#705 is on https://build.ros2.org, so I'm wondering why there is a difference there.  Could it maybe have something to do with installing Connext from the deb package, vs. installing a different way?

ivanpauno:
> Could it maybe have something to do with installing Connext from the deb package, vs. installing a different way?

I thought it was because of the update to Connext 6 (that already happened, right?), but maybe it was actually a difference between the deb package vs the default way CI uses.
When I installed ROS 2 rolling including Connext from debians, I see that the warning about the license is:

```
RTI Connext DDS Non-commercial license is for academic, research, evaluation and personal use only. USE FOR COMMERCIAL PURPOSES IS PROHIBITED. See RTI_LICENSE.TXT for terms. Download free tools at rti.com/ncl. License issued to Non-Commercial User license@rti.com For non-production use only.
Expires on 00-jan-00 See www.rti.com for more information.
```

Which starts with `RTI Data Distribution Service`, not with `RTI Connext DDS`.
That also matches the error output reported in https://github.com/ros2/ros2cli/issues/705.
My development environment is still running in Ubuntu 18.04, in that case I'm using Connext 5.3 installed from debians (`rti-connext-dds-5.3.1`), and the message starts with `RTI Connext DDS` (which was what the old output filter was ignoring correctly).

So maybe the license warning when Connext 6 is installed from debians or the other way we're using in CI (I'm not sure what that is) have different license warning messages (?).

Running CI again, this time using connext debs:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/16534/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/11121/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/16908/)

If that passes, I guess it should solve the issue in build.ros2.org as well.

clalancette:
> I thought it was because of the update to Connext 6 (that already happened, right?)

Yes, that has already happened.  It *should* have happened both on https://build.ros2.org and on https://ci.ros2.org, but there may be differences in how they are installed.  Let's see if your new CI shows anything interesting.

ivanpauno:
I see that the jobs that are running above have already some tests failing, but I forgot to specify the branch :man_facepalming: .
Anyway, the fact that it's failing now with connext debs is good, it seems that actually the license message warning changes depending on the installation method.

Running CI again, this time using the correct branch:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/16535/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/11122/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/16909/)

clalancette:
Ah, so I guess there is a difference on how they are installed.  OK, in that case this makes a lot more sense.

clalancette (approved):

PR #84, opened 2022-04-22
Handle 'best_available' QoS policies

j-rivero:
Connects to: https://github.com/ros2/rmw/pull/320
Twin of: https://github.com/ros2/rmw_fastrtps/pull/598

If users set certain QoS policies to BEST_AVAILABLE, then we query the graph for endpoint info. The policy is updated such that it is compatible with all endpoints while maintaining the highest possible level of service.

### DIFF
--- a/rmw_connextdds/src/rmw_api_impl_ndds.cpp
+++ b/rmw_connextdds/src/rmw_api_impl_ndds.cpp
@@ -12,6 +12,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+#include "rmw_dds_common/qos.hpp"
+
 #include "rmw_connextdds/rmw_api_impl.hpp"
 
 /*****************************************************************************
@@ -377,8 +379,18 @@ rmw_create_publisher(
   const rmw_qos_profile_t * qos_policies,
   const rmw_publisher_options_t * publisher_options)
 {
+  RMW_CHECK_ARGUMENT_FOR_NULL(qos_policies, nullptr);
+
+  // Adapt any 'best available' QoS options
+  rmw_qos_profile_t adapted_qos_policies = *qos_policies;
+  rmw_ret_t ret = rmw_dds_common::qos_profile_get_best_available_for_topic_publisher(
+    node, topic_name, &adapted_qos_policies, rmw_get_subscriptions_info_by_topic);
+  if (RMW_RET_OK != ret) {
+    return nullptr;
+  }
+
   return rmw_api_connextdds_create_publisher(
-    node, type_supports, topic_name, qos_policies, publisher_options);
+    node, type_supports, topic_name, &adapted_qos_policies, publisher_options);
 }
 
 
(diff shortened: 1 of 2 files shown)


rmw_connextdds/src/rmw_api_impl_ndds.cpp
```diff
@@ -694,8 +706,18 @@ rmw_create_subscription(
...
+
+  // Adapt any 'best available' QoS options
+  rmw_qos_profile_t adapted_qos_policies = *qos_policies;
+  rmw_ret_t ret = rmw_dds_common::qos_profile_get_best_available_for_topic_subscription(
+    node, topic_name, &adapted_qos_policies, rmw_get_publishers_info_by_topic);
+  if (RMW_RET_OK != ret) {
+    return nullptr;
+  }
+
   return rmw_api_connextdds_create_subscription(
```
jacobperron, line 719: Maybe we can move the new logic into [rmw_api_connextdds_create_subscription](https://github.com/ros2/rmw_connextdds/blob/24c86f5a24dfa5c56730bf2e99892b8550e0f29c/rmw_connextdds_common/src/common/rmw_subscription.cpp#L50)? Then we don't have to repeat ourselves both here and in rmw_connextddsmicro. Same idea for the create publisher logic.

jacobperron: Actually, maybe we can't make it work because of a linking issue. The function pointer passed to the rmw_dds_common function is technically implemented in this package, not the common package. I had the same issue in the connected rmw_fastrtps PR.

jacobperron (approved):

rmw_connextdds/src/rmw_api_impl_ndds.cpp
asorbini: I'm not sure I understand the possible problem, but I agree with @jacobperron that this logic should definitely be moved to `rmw_api_connextdds_create_subscription()` (and similarly for publishers, clients, and services), so that it may be shared between both RMWs.

asorbini (changes requested):

rmw_connextdds/src/rmw_api_impl_ndds.cpp
j-rivero: I coded the approach to give a try on building and testing. Worked fine in my local focal system until the point of compiling the system_tests. See #85

asorbini: I see now, thank you for clarifying @jacobperron. There are `rmw_api_connextdds_get_publishers_info_by_topic` and `rmw_api_connextdds_get_subscriptions_info_by_topic` which provide common implementations in `rmw_connextdds_common`.

@j-rivero #85 looks like the right approach, but it's missing the publisher part (which should go within `rmw_api_connextdds_create_publisher` in `rmw_publication.cpp`)

j-rivero: Thanks Andrea, I've implemented your recommendation but test are still failing. Let's continue there https://github.com/ros2/rmw_connextdds/pull/85#issuecomment-1112382989

PR #96, opened 2022-12-20
Add in implementation of inconsistent topic.

clalancette:
Signed-off-by: Chris Lalancette <clalancette@openrobotics.org>

Part of https://github.com/ros2/ros2/issues/1361

### DIFF
--- a/rmw_connextdds_common/include/rmw_connextdds/rmw_waitset_std.hpp
+++ b/rmw_connextdds_common/include/rmw_connextdds/rmw_waitset_std.hpp
@@ -243,8 +243,7 @@ public:
   explicit RMW_Connext_StatusCondition(
     DDS_Entity * const entity)
   : scond(DDS_Entity_get_statuscondition(entity)),
-    status_inconsistent_topic(DDS_InconsistentTopicStatus_INITIALIZER),
-    status_inconsistent_topic_last(DDS_InconsistentTopicStatus_INITIALIZER)
+    status_inconsistent_topic(DDS_InconsistentTopicStatus_INITIALIZER)
   {
     this->scond = DDS_Entity_get_statuscondition(entity);
     if (nullptr == this->scond) {
@@ -351,7 +350,6 @@ public:
 
         this->triggered_inconsistent_topic = false;
         this->status_inconsistent_topic.total_count_change = 0;
-        this->status_inconsistent_topic_last = this->status_inconsistent_topic;
       }, false /* notify */);
 
     return RMW_RET_OK;
@@ -363,8 +361,6 @@ protected:
   bool triggered_inconsistent_topic{false};
 
   struct DDS_InconsistentTopicStatus status_inconsistent_topic;
-
-  struct DDS_InconsistentTopicStatus status_inconsistent_topic_last;
 };
 
 void
--- a/rmw_connextdds_common/include/rmw_connextdds/rmw_waitset_std.hpp
+++ b/rmw_connextdds_common/include/rmw_connextdds/rmw_waitset_std.hpp
@@ -242,7 +242,9 @@ class RMW_Connext_StatusCondition : public RMW_Connext_Condition
 public:
   explicit RMW_Connext_StatusCondition(
     DDS_Entity * const entity)
-  : scond(DDS_Entity_get_statuscondition(entity))
+  : scond(DDS_Entity_get_statuscondition(entity)),
+    status_inconsistent_topic(DDS_InconsistentTopicStatus_INITIALIZER),
+    status_inconsistent_topic_last(DDS_InconsistentTopicStatus_INITIALIZER)
   {
     this->scond = DDS_Entity_get_statuscondition(entity);
     if (nullptr == this->scond) {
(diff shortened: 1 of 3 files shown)


rmw_connextdds_common/include/rmw_connextdds/rmw_waitset_std.hpp
```diff
@@ -333,8 +335,38 @@ class RMW_Connext_StatusCondition : public RMW_Connext_Condition
...
+    rmw_inconsistent_topic_status_t * const status)
+  {
+    update_state(
+      [this, status]() {
+        this->triggered_inconsistent_topic = false;
+
+        status->total_count = this->status_inconsistent_topic.total_count;
+        status->total_count_change = this->status_inconsistent_topic.total_count_change;
+
+        this->status_inconsistent_topic.total_count_change = 0;
```
fujitatomoya, line 354: Do we need to clear this variable with `0`? if the next time `total_count_change` is `0`, it should be also updated by `rticonnextdds`?

rmw_connextdds_common/include/rmw_connextdds/rmw_waitset_std.hpp
```diff
@@ -333,8 +335,38 @@ class RMW_Connext_StatusCondition : public RMW_Connext_Condition
...
+  }
+
 protected:
   DDS_StatusCondition * scond;
+
+  bool triggered_inconsistent_topic{false};
+
+  struct DDS_InconsistentTopicStatus status_inconsistent_topic;
+
+  struct DDS_InconsistentTopicStatus status_inconsistent_topic_last;
```
fujitatomoya, line 369: Do we need to keep the last status? Why do we not just store the current status from RTI connextdds to RMW?

rmw_connextdds_common/src/common/rmw_impl_waitset_std.cpp
```diff
@@ -550,6 +564,28 @@ RMW_Connext_WaitSet::wait(
...
+void
+RMW_Connext_StatusCondition::update_status_inconsistent_topic(
+  const struct DDS_InconsistentTopicStatus * status)
+{
+  this->status_inconsistent_topic = *status;
+  this->triggered_inconsistent_topic = true;
+
+  this->status_inconsistent_topic.total_count_change = this->status_inconsistent_topic.total_count;
+  this->status_inconsistent_topic.total_count_change -=
+    this->status_inconsistent_topic_last.total_count;
```
fujitatomoya, line 586: I am not sure if this calculation is required to be done in RMW, it seems that we can just store the data from DDS w/o processing?

besides, `total_count_change` also needs to count `undiscovered` inconsistent topic? i was thinking that it just incremental number to `discover` inconsistent topics only.

fujitatomoya (commented):
@clalancette a few questions, i am not really sure what the correct behavior is.

rmw_connextdds_common/include/rmw_connextdds/rmw_waitset_std.hpp
clalancette: Honestly, I'm not sure.  I pretty much copied this from https://github.com/ros2/rmw_connextdds/blob/806fdfb6e22b587c850bdf153c12a448c6b19e3d/rmw_connextdds_common/include/rmw_connextdds/rmw_waitset_std.hpp#L454

rmw_connextdds_common/src/common/rmw_impl_waitset_std.cpp
clalancette: Let me test this out and see if we can just copy it out of status directly.

rmw_connextdds_common/include/rmw_connextdds/rmw_waitset_std.hpp
clalancette: And this is now removed in 8f7660e

rmw_connextdds_common/src/common/rmw_impl_waitset_std.cpp
clalancette: And yeah, we can.  So this is removed in 8f7660eefd6bb159944e9881fe1459e947e23d49

mjcarroll (approved):

rmw_connextdds_common/include/rmw_connextdds/rmw_waitset_std.hpp
```diff
@@ -333,8 +334,33 @@ class RMW_Connext_StatusCondition : public RMW_Connext_Condition
   virtual bool
   has_status(const rmw_event_type_t event_type) = 0;
 
+  void
+  on_inconsistent_topic(const struct DDS_InconsistentTopicStatus * status);
+
+  void update_status_inconsistent_topic(const struct DDS_InconsistentTopicStatus * status);
```
methylDragon, line 340: Nit

```suggestion
  void
  update_status_inconsistent_topic(const struct DDS_InconsistentTopicStatus * status);
```

methylDragon (approved):
LGTM on green CI, with a minor nit

I won't run CI here since it's run in the tracking issue

rmw_connextdds_common/include/rmw_connextdds/rmw_waitset_std.hpp
clalancette: Good spot, fixed in 44fe0a6

PR #106, opened 2023-02-27
get valid subscription handle even if cft failed

iuhilnehc-ynos:
rcl need to continue to try rcl cft fallback if rmw implementation failed to call rmw cft.

### DIFF
--- a/rmw_connextdds_common/src/common/rmw_impl.cpp
+++ b/rmw_connextdds_common/src/common/rmw_impl.cpp
@@ -1296,6 +1296,9 @@ RMW_Connext_Subscriber::create(
 
   if (RMW_RET_OK == cft_rc) {
     sub_topic = cft_topic;
+  } else {
+    // RMW_Connext_Subscriber needs this value to check if CFT is enabled or not.
+    sub_cft_expr = "";
   }
 
   // The following initialization generates warnings when built
--- a/rmw_connextdds_common/src/common/rmw_impl.cpp
+++ b/rmw_connextdds_common/src/common/rmw_impl.cpp
@@ -1294,11 +1294,7 @@ RMW_Connext_Subscriber::create(
     sub_cft_params,
     &cft_topic);
 
-  if (RMW_RET_OK != cft_rc) {
-    if (RMW_RET_UNSUPPORTED != cft_rc) {
-      return nullptr;
-    }
-  } else {
+  if (RMW_RET_OK == cft_rc) {
     sub_topic = cft_topic;
   }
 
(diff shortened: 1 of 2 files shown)


rmw_connextdds_common/src/common/rmw_impl.cpp
```diff
@@ -1294,12 +1294,11 @@ RMW_Connext_Subscriber::create(
...
-  if (RMW_RET_OK != cft_rc) {
-    if (RMW_RET_UNSUPPORTED != cft_rc) {
-      return nullptr;
-    }
-  } else {
+  if (RMW_RET_OK == cft_rc) {
     sub_topic = cft_topic;
+  } else {
+    // RMW_Connext_Subscriber needs this value to check if CFT is enabled or not.
+    sub_cft_expr = "";
```
fujitatomoya, line 1301: this is required for `RMW_Connext_Subscriber` w/o content_fitlered_topic, right?

fujitatomoya (approved):
one comment needs to be confirmed but lgtm.

fujitatomoya:
@asorbini we would like to request the another review on this.

asorbini (commented):
@iuhilnehc-ynos thank you for submitting these changes, and apologies for the late reply.

The change looks good to me, but I'm not sure why they are needed.

Is the ROS filtering syntax different than the standard DDS one?

If not, I would assume a failure to create the CFT signals some problem with the parameters, and it should not be ignored.

If there are filters that are only supported by the rcl implementation, then this change is fine.

rmw_connextdds_common/src/common/rmw_impl.cpp
iuhilnehc-ynos: Yeah, RMW_Connext_Subscriber need this variable https://github.com/ros2/rmw_connextdds/blob/5d079a44112adce70451b26c6193b5de7fcbfad2/rmw_connextdds_common/src/common/rmw_impl.cpp#L1370, and set it into [cft_expression](https://github.com/ros2/rmw_connextdds/blob/5d079a44112adce70451b26c6193b5de7fcbfad2/rmw_connextdds_common/src/common/rmw_impl.cpp#L1148), which is used in https://github.com/ros2/rmw_connextdds/blob/5d079a44112adce70451b26c6193b5de7fcbfad2/rmw_connextdds_common/include/rmw_connextdds/rmw_impl.hpp#L490

iuhilnehc-ynos:
@asorbini 

Thanks for your reply. No matter when you reply, it's never too late.

> Is the ROS filtering syntax different than the standard DDS one?

No, I think it's the same for now, maybe `RCL content filter fallback` can support extra syntax in the future.
The extra reason for patching this is the `RCL content filter fallback` could be used if rmw_connextdds is unsupported while using the `rtime` version.

fujitatomoya:
see https://github.com/ros2/demos/pull/579#issuecomment-3757278811

PR #108, opened 2023-03-10
[rmw_connextdds] New RMW discovery options

mxgrey:
This PR updates `rmw_connextdds` to be compatible with the upcoming [discovery options feature for RMW](https://discourse.ros.org/t/proposed-changes-to-how-ros-performs-discovery-of-nodes/27640). (Note that the final agreed upon requirements matrix is [the one in this comment](https://discourse.ros.org/t/proposed-changes-to-how-ros-performs-discovery-of-nodes/27640/25)).

This PR depends on these upstream PRs:
* https://github.com/ros2/rmw/pull/349
* https://github.com/ros2/rcl/pull/1038

With this PR, the requirements matrix for the new discovery options is *almost* fully satisfied. There is just one aspect that is not working correctly, which I would appreciate some help with from Connext DDS experts: **It seems that Connext DDS participants are not accepting unknown initial peers even though `accept_unknown_peers` is being set to `true`.** As a result, the `Same Host` matrix is 100% passing while the `Other Host` matrix looks like this:

* :ok:: Required behavior is passing
* :red_circle:: Required behavior is failing

Different Hosts |   |   | Node B setting |   |   |   |   |  
-- | -- | -- | -- | -- | -- | -- | -- | --
  |   |   | No static peer |   |   | With static peer |   |  
  |   |   | Off | Localhost | Subnet | Off | Localhost | Subnet
Node A setting | No static peer | Off | :ok: | :ok: | :ok: | :ok: | :ok: | :ok:
  |   | Localhost | :ok: | :ok: | :ok: | :ok: | :red_circle: | :red_circle:
  |   | Subnet | :ok: | :ok: | :ok: | :ok: | :red_circle: | :ok:
  | With static peer | Off | :ok: | :ok: | :ok: | :ok: | :ok: | :ok:
  |   | Localhost | :ok: | :red_circle: | :red_circle: | :ok: | :ok: | :ok: 
  |   | Subnet | :ok:  | :red_circle: | :ok: | :ok: | :ok: | :ok: 

The gist of the current problem is: 
* when two endpoints are on different hosts 
* and one or both endpoints are in `LOCALHOST` mode
* and neither endpoint is in `OFF` mode
* and only one endpoint declares the other as a static peer

-> then the required behavior is for them to discover each other and connect, but they currently do not. Every other scenario is working correctly.

From my reading of the Connext DDS documentation, I believe the current implementation is supposed to work since `accept_unknown_peers` is set to true, but perhaps there is a detail that I'm overlooking.

One observation from watching packets with wireshark is that when there's a `LOCALHOST + static peer <=> LOCALHOST + no static peer` connection, the endpoint that has a static peer is sending `DATA(p)` participant information packets while the endpoint without a static peer is never sending the `DATA(p)` packets (..even though it does send various ACKNACK and HEARTBEAT packets). I'm happy to provide the wireshark data if that would be helpful.

### DIFF
--- a/rmw_connextdds_common/src/common/rmw_context.cpp
+++ b/rmw_connextdds_common/src/common/rmw_context.cpp
@@ -25,6 +25,8 @@
 
 #include "rcutils/env.h"
 #include "rcutils/filesystem.h"
+#include "rcutils/process.h"
+#include "rcutils/snprintf.h"
 
 /******************************************************************************
  * Global reference to the Domain Participant Factory.
@@ -75,6 +77,185 @@ rmw_connextdds_initialize_participant_factory_qos(
 }
 
 static
+rmw_ret_t rmw_connextdds_extend_initial_peer_list(
+  const rmw_peer_address_t * const static_peers,
+  const size_t static_peer_count,
+  struct DDS_StringSeq * const out)
+{
+  if (static_peer_count == 0) {
+    return RMW_RET_OK;
+  }
+
+  if (static_peers == nullptr) {
+    RMW_CONNEXT_LOG_ERROR_A_SET(
+      "received nullptr static_peers but a static_peer_count of %lu",
+      static_peer_count);
+    return RMW_RET_ERROR;
+  }
+
+  const auto initial_length = DDS_StringSeq_get_length(out);
+  const auto new_seq_length = initial_length + static_peer_count;
+  if (!DDS_StringSeq_ensure_length(out, new_seq_length, new_seq_length)) {
+    RMW_CONNEXT_LOG_ERROR_SET("failed to resize string sequence")
+    return RMW_RET_ERROR;
+  }
+  for (size_t s = 0; s < static_peer_count; ++s) {
+    const auto index = initial_length + s;
+    const char * peer = static_peers[s].peer_address;
+    char ** const element_ref = DDS_StringSeq_get_reference(out, index);
+    RMW_CONNEXT_ASSERT(nullptr != element_ref);
+    if (nullptr != *element_ref) {
+      /* If some strings are still hanging around this dead space in the seq,
+         then free their memory. */
+      DDS_String_free(*element_ref);
+    }
+    *element_ref = DDS_String_dup(peer);
+    if (nullptr == *element_ref) {
+      RMW_CONNEXT_LOG_ERROR_A_SET("failed to duplicate peer string: %s", peer);
+      return RMW_RET_ERROR;
+    }
+
+    RMW_CONNEXT_LOG_TRACE_A(
+      "inserted static peer: i=%d, peer='%s'",
+      index, peer);
+  }
+
+  return RMW_RET_OK;
+}
+
+static
+rmw_ret_t
+rmw_connextdds_initialize_discovery_options(
+  rmw_context_impl_t * const ctx,
+  DDS_DomainParticipantQos & dp_qos)
+{
+  const auto range = ctx->discovery_options->automatic_discovery_range;
+  switch (range) {
+    case RMW_AUTOMATIC_DISCOVERY_RANGE_SYSTEM_DEFAULT:
+    case RMW_AUTOMATIC_DISCOVERY_RANGE_SUBNET:
+      /* No action needed. This is the default discovery behavior for DDS */
+      break;
+    case RMW_AUTOMATIC_DISCOVERY_RANGE_NOT_SET:
+    case RMW_AUTOMATIC_DISCOVERY_RANGE_LOCALHOST:
+    case RMW_AUTOMATIC_DISCOVERY_RANGE_OFF:
+      /*Limit the UDPv4 transport to use only unicast interfaces, by setting
+        the list of multicast interfaces to a "localhost" value.
+        Note: We allow the LOCALHOST interface for the OFF range
+        because if we leave this property completely blank then it
+        has the opposite effect and allows all interfaces to be used.
+        Allowing only LOCALHOST at least minimizes the unnecessary
+        discovery traffic and prevents discovery with other host
+        machines, while the domain_tag protects against same-host
+        connections. */
+      if (DDS_RETCODE_OK != DDS_PropertyQosPolicyHelper_assert_property(
+          &dp_qos.property,
+          "dds.transport.UDPv4.builtin.parent.allow_multicast_interfaces_list",
+          RMW_CONNEXT_LOCALHOST_ONLY_ADDRESS,
+          DDS_BOOLEAN_FALSE /* propagate */))
+      {
+        RMW_CONNEXT_LOG_ERROR_A_SET(
+          "failed to assert property on participant: %s",
+          "dds.transport.UDPv4.builtin.parent.allow_multicast_interfaces_list")
+        return RMW_RET_ERROR;
+      }
+      break;
+    default:
+      RMW_CONNEXT_LOG_ERROR_A_SET(
+        "Unknown value provided for automatic discovery range: %i",
+        range);
+      return RMW_RET_ERROR;
+  }
+
+  if (RMW_AUTOMATIC_DISCOVERY_RANGE_OFF == range) {
+    // When discovery rage is RMW_AUTOMATIC_DISCOVERY_RANGE_OFF,
+    // prevent the participant from discovery anyone by setting an empty
+    // initial peers list and disabling "accept_unknown_peers".
+    // Also, assign a host-wide unique domain tag to the participant to
+    // prevent discovery with other local participant (e.g. through shared
+    // memory transport).
+    const DDS_Long ros_peers = DDS_StringSeq_get_length(&ctx->initial_peers);
+    const DDS_Long qos_peers = DDS_StringSeq_get_length(&dp_qos.discovery.initial_peers);
+    dp_qos.discovery.accept_unknown_peers = DDS_BOOLEAN_FALSE;
+    if (ros_peers > 0) {
+      RMW_CONNEXT_LOG_WARNING_A(
+        "requested %d initial peers using %s, but discovery range is off",
+        ros_peers,
+        RMW_CONNEXT_ENV_INITIAL_PEERS);
+      if (!DDS_StringSeq_ensure_length(&ctx->initial_peers, 0, 0)) {
+        RMW_CONNEXT_LOG_ERROR_SET("failed to clear initial peers list")
+        return RMW_RET_ERROR;
+      }
+    }
+    if (qos_peers > 0) {
+      RMW_CONNEXT_LOG_WARNING_A(
+        "requested %d initial peers from DomainParticipantQos, but discovery range is off",
+        qos_peers);
+      if (!DDS_StringSeq_ensure_length(&ctx->initial_peers, 0, 0)) {
+        RMW_CONNEXT_LOG_ERROR_SET("failed to clear initial peers list")
+        return RMW_RET_ERROR;
+      }
+    }
+    /* See earlier note about why we allow LOCALHOST interface for
+        OFF range. */
+    if (DDS_RETCODE_OK != DDS_PropertyQosPolicyHelper_assert_property(
+        &dp_qos.property,
+        "dds.transport.UDPv4.builtin.parent.allow_interfaces_list",
+        RMW_CONNEXT_LOCALHOST_ONLY_ADDRESS,
+        DDS_BOOLEAN_FALSE /* propagate */))
+    {
+      RMW_CONNEXT_LOG_ERROR_SET(
+        "failed to assert property on participant: "
+        "dds.transport.UDPv4.builtin.parent.allow_interfaces_list");
+      return RMW_RET_ERROR;
+    }
+
+    /* Give this participant its own unique domain tag to prevent
+        unicast discovery from happening. */
+    if (!ctx->domain_tag) {
+      const auto pid = rcutils_get_pid();
+      static const char * format_string = "ros_discovery_off_%d";
+      const int bytes_needed = rcutils_snprintf(nullptr, 0, format_string, pid);
+      ctx->domain_tag = DDS_String_alloc(bytes_needed);
+      if (nullptr == ctx->domain_tag) {
+        RMW_CONNEXT_LOG_ERROR_SET("failed to allocate domain tag string");
+        return RMW_RET_BAD_ALLOC;
+      }
+      if (rcutils_snprintf(ctx->domain_tag, bytes_needed + 1, format_string, pid) < 0) {
+        RMW_CONNEXT_LOG_ERROR_SET("failed to format ros discovery off information into domain tag");
+        return RMW_RET_ERROR;
+      }
+    }
+    if (DDS_RETCODE_OK != DDS_PropertyQosPolicyHelper_assert_property(
+        &dp_qos.property,
+        "dds.domain_participant.domain_tag",
+        ctx->domain_tag,
+        DDS_BOOLEAN_FALSE))
+    {
+      RMW_CONNEXT_LOG_ERROR_SET(
+        "failed to assert property on participant: "
+        "dds.domain_participant.domain_tag");
+      return RMW_RET_ERROR;
+    }
+  } else {
+    // For any other discovery range, copy the list of static peers to so that
+    // it will be later copied to DomainParticipantQos::discovery::initial_peers.
+    dp_qos.discovery.accept_unknown_peers = DDS_BOOLEAN_TRUE;
+    const auto rc = rmw_connextdds_extend_initial_peer_list(
+      ctx->discovery_options->static_peers,
+      ctx->discovery_options->static_peers_count,
+      &ctx->initial_peers);
+    if (RMW_RET_OK != rc) {
+      RMW_CONNEXT_LOG_ERROR(
+        "failed to extend initial peers with the static peers");
+      return rc;
+    }
+  }
+
+  return RMW_RET_OK;
+}
+
+
+static
 rmw_ret_t
 rmw_connextdds_initialize_participant_qos(
   rmw_context_impl_t * const ctx,
@@ -97,20 +278,22 @@ rmw_connextdds_initialize_participant_qos(
   switch (ctx->participant_qos_override_policy) {
     case rmw_context_impl_t::participant_qos_override_policy_t::All:
     case rmw_context_impl_t::participant_qos_override_policy_t::Basic:
-      {
-        if (DDS_StringSeq_get_length(&ctx->initial_peers) > 0)
-        {
-          if (!DDS_StringSeq_copy(&dp_qos.discovery.initial_peers, &ctx->initial_peers)) {
-            RMW_CONNEXT_LOG_ERROR_SET("failed to copy initial peers sequence")
-            return RMW_RET_ERROR;
-          }
+      if (nullptr != ctx->discovery_options) {
+        const auto rc = rmw_connextdds_initialize_discovery_options(ctx, dp_qos);
+        if (RMW_RET_OK != rc) {
+          RMW_CONNEXT_LOG_ERROR("failed to initialize discovery options")
+          return RMW_RET_ERROR;
         }
-        break;
       }
-    default:
+      if (DDS_StringSeq_get_length(&ctx->initial_peers) > 0 &&
+        !DDS_StringSeq_copy(&dp_qos.discovery.initial_peers, &ctx->initial_peers))
       {
-        break;
+        RMW_CONNEXT_LOG_ERROR_SET("failed to copy initial peers sequence")
+        return RMW_RET_ERROR;
       }
+      break;
+    default:
+      break;
   }
 
   return RMW_RET_OK;
@@ -122,7 +305,9 @@ rmw_context_impl_t::initialize_node(
 {
   if (this->node_count > 0) {
     bool params_equal = false;
-    if (rmw_discovery_options_equal(this->discovery_options, discovery_options, &params_equal) != RMW_RET_OK) {
+    if (rmw_discovery_options_equal(
+        this->discovery_options, discovery_options, &params_equal) != RMW_RET_OK)
+    {
       RMW_CONNEXT_LOG_ERROR_SET("invalid discovery params argument");
       return RMW_RET_INVALID_ARGUMENT;
     }
@@ -135,10 +320,10 @@ rmw_context_impl_t::initialize_node(
   } else {
     if (nullptr != discovery_options) {
       RMW_CONNEXT_ASSERT(nullptr == this->discovery_options)
-      this->discovery_options = (rmw_discovery_options_t*)
+      this->discovery_options = static_cast<rmw_discovery_options_t *>(
         this->base->options.allocator.allocate(
-        sizeof(rmw_discovery_options_t),
-        this->base->options.allocator.state);
+          sizeof(rmw_discovery_options_t),
+          this->base->options.allocator.state));
       if (nullptr == this->discovery_options) {
         RMW_CONNEXT_LOG_ERROR_SET("failed to allocate discovery options")
         return RMW_RET_ERROR;
@@ -167,30 +352,30 @@ rmw_context_impl_t::initialize_node(
       if (RMW_AUTOMATIC_DISCOVERY_RANGE_LOCALHOST ==
         this->discovery_options->automatic_discovery_range)
       {
-        static const char * const NDDS_DISCOVERY_PEERS = "NDDS_DISCOVERY_PEERS";
+        static const char * const ndds_disc_peers_var = "NDDS_DISCOVERY_PEERS";
         const char * ndds_disc_peers = nullptr;
-        const char * lookup_rc = rcutils_get_env(NDDS_DISCOVERY_PEERS, &ndds_disc_peers);
+        const char * lookup_rc = rcutils_get_env(ndds_disc_peers_var, &ndds_disc_peers);
         if (nullptr != lookup_rc || nullptr == ndds_disc_peers) {
           RMW_CONNEXT_LOG_ERROR_A_SET(
             "failed to lookup from environment: "
             "var=%s, "
             "rc=%s ",
-            NDDS_DISCOVERY_PEERS,
+            ndds_disc_peers_var,
             lookup_rc)
           return RMW_RET_ERROR;
         }
-        if (ndds_disc_peers[0] == '\0'
-          && !rcutils_set_env(NDDS_DISCOVERY_PEERS, RMW_CONNEXT_LOCALHOST_ONLY_ADDRESS))
+        if (ndds_disc_peers[0] == '\0' &&
+          !rcutils_set_env(ndds_disc_peers_var, RMW_CONNEXT_LOCALHOST_ONLY_ADDRESS))
         {
           RMW_CONNEXT_LOG_ERROR_A_SET(
-            "failed to set environment: var=%s",
-            NDDS_DISCOVERY_PEERS)
+            "failed to set environment: var=%s", ndds_disc_peers_var)
           return RMW_RET_ERROR;
         }
       }
     }
 
     if (nullptr == RMW_Connext_gv_DomainParticipantFactory) {
+      RMW_CONNEXT_ASSERT(1 == RMW_Connext_gv_ContextCount)
       RMW_CONNEXT_LOG_DEBUG("initializing DDS DomainParticipantFactory")
 
       if (RMW_RET_OK !=
--- a/rmw_connextdds_common/src/common/rmw_context.cpp
+++ b/rmw_connextdds_common/src/common/rmw_context.cpp
@@ -249,6 +249,19 @@ rmw_connextdds_initialize_discovery_options(
         "failed to extend initial peers with the static peers");
       return rc;
     }
+    /* Support for ROS_AUTOMATIC_DISCOVERY_RANGE == LOCALHOST
+      -----------------------------------------
+      Make sure that the participant is not listening on any multicast address. */
+    if (RMW_AUTOMATIC_DISCOVERY_RANGE_LOCALHOST ==
+      ctx->discovery_options->automatic_discovery_range)
+    {
+      if (!DDS_StringSeq_ensure_length(
+          &dp_qos.discovery.multicast_receive_addresses, 0, 0))
+      {
+        RMW_CONNEXT_LOG_ERROR_SET("failed to resize multicast_receive_addresses")
+        return RMW_RET_ERROR;
+      }
+    }
   }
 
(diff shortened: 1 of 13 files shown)


rmw_connextdds_common/src/common/rmw_context.cpp
```diff
@@ -109,35 +109,44 @@ rmw_connextdds_initialize_participant_qos(
...
     this->node_count += 1;
     RMW_CONNEXT_LOG_DEBUG_A(
-      "initialized new node: total=%lu", this->node_count)
+      "initialized new node: total=%lu", this->node_count);
     return RMW_RET_OK;
   }
 
-  rmw_ret_t rc = this->initialize_participant(localhost_only);
+  if (!this->discovery_options && discovery_options) {
+    this->discovery_options = (rmw_discovery_options_t*)
```
wjwwood, line 134: This is C++ so, probably should use a C++-style cast here.

rmw_connextdds_common/src/common/rmw_context.cpp
```diff
@@ -109,35 +109,44 @@ rmw_connextdds_initialize_participant_qos(
...
+      this->base->options.allocator.allocate(
+      sizeof(rmw_discovery_options_t),
+      this->base->options.allocator.state);
+
+    const auto rc = rmw_discovery_options_copy(
+      discovery_options,
+      this->base->options.allocator,
+      this->discovery_options);
+    if (RMW_RET_OK != rc) {
+      RMW_CONNEXT_LOG_ERROR("failed to copy discovery parameters");
```
wjwwood, line 144: This should check the error message and add it to the log message. Or better yet, add to the error message and return the return code. Unconditionally logging here doesn't seem like exactly the right thing to do.

rmw_connextdds_common/src/common/rmw_context.cpp
```diff
@@ -991,7 +1013,7 @@ rmw_api_connextdds_init(
   ctx->optimize_large_data = '\0' == disable_optimize_large_data_env[0];
 #endif /* RMW_CONNEXT_DEFAULT_LARGE_DATA_OPTIMIZATIONS */
 
-  /* Lookup and configure initial peer from environment */
+    /* Lookup and configure initial peer from environment */
```
wjwwood, line 1016: ?

rmw_connextdds_common/src/common/rmw_impl.cpp
```diff
@@ -180,7 +180,7 @@ rmw_connextdds_parse_string_list(
       DDS_String_free(*el_ref);
     }
     *el_ref = DDS_String_alloc(next_len);
-    if (nullptr == el_ref) {
+    if (nullptr == *el_ref) {
```
wjwwood, line 183: This could be right...? I want to double check because it seems unrelated to the changes for this feature.

rmw_connextdds_common/src/common/rmw_impl.cpp
```diff
@@ -205,6 +205,51 @@ rmw_connextdds_parse_string_list(
...
+      DDS_String_free(*element_ref);
+    }
+    const auto peer_char_len = strlen(peer);
+    *element_ref = DDS_String_alloc(peer_char_len);
+    if (nullptr == *element_ref) {
+      RMW_CONNEXT_LOG_ERROR_SET("failed to allocate string");
+      return RMW_RET_ERROR;
+    }
+
+    memcpy(*element_ref, peer, peer_char_len);
```
wjwwood, line 244: I always get nervous when I see a combination of strlen and memcpy, since it can be easy to accidentally not copy the null terminator. I think that might actually be the case here, but I'm not 100% sure.

https://stackoverflow.com/questions/8262406/get-strlen-after-memcpy-in-c

rmw_connextdds_common/src/ndds/dds_api_ndds.cpp
```diff
@@ -197,25 +200,119 @@ rmw_connextdds_initialize_participant_qos_impl(
...
+            case RMW_AUTOMATIC_DISCOVERY_RANGE_SUBNET:
+              /* No action needed. This is the default discovery behavior for DDS */
+              break;
+            default:
+              RMW_CONNEXT_LOG_WARNING_A(
+                "Unknown value provided for automatic discovery range: %i",
+                ctx->discovery_options->automatic_discovery_range);
+              /* Fall back to the default behavior */
+              [[clang::fallthrough]];
+              /* [[fallthrough]]; // Uncomment this when migrating to C++17 */
```
wjwwood, line 228: Same comment as on the other pr, we have access to C++17, this should either be removed or addressed I think.

rmw_connextdds_common/src/ndds/dds_api_ndds.cpp
```diff
@@ -197,25 +200,119 @@ rmw_connextdds_initialize_participant_qos_impl(
   dp_qos->user_object.topic_user_object.size = sizeof(void *);
   dp_qos->user_object.content_filtered_topic_user_object.size = sizeof(void *);
 #endif /* RMW_CONNEXT_SHARE_DDS_ENTITIES_WITH_CPP */
+  if (!ctx->domain_tag) {
+    const auto pid = rcutils_get_pid();
+    ctx->domain_tag = DDS_String_alloc(18 + log10(pid));
+    sprintf(ctx->domain_tag, "ros_discovery_off_%d", pid);
```
wjwwood, line 206: This would feel much safer to me if we just created the string we want, then take the size of it to allocate into the domain tag, because if someone updated the string `"ros_discovery..."` whatever to something else, and forgot to update the alloc line above... that would be bad and it would be easy to do.

asorbini: I agree, the string should be declared as a static constant and the length should not be a "magic number".
Also, the return value for `DDS_String_alloc()` must be checked.

```suggestion
    static const std::string domain_tag_pfx = "ros_discovery_off_";
    const auto pid = rcutils_get_pid();
    const auto domain_tag_len = domain_tag_pfx.length() + log10(pid);
    ctx->domain_tag = DDS_String_alloc(domain_tag_len);
    if (nullptr == ctx->domain_tag) {
      RMW_CONNEXT_LOG_ERROR_SET("failed to allocate domain tag string");
      return RMW_RET_ERROR;
    }
    sprintf(ctx->domain_tag, "%s%d", domain_tag_pfx.c_str(), pid);
```

rmw_connextdds_common/src/ndds/dds_api_ndds.cpp
```diff
@@ -197,25 +200,119 @@ rmw_connextdds_initialize_participant_qos_impl(
...
+                "dds.transport.UDPv4.builtin.parent.allow_multicast_interfaces_list",
+                RMW_CONNEXT_LOCALHOST_ONLY_ADDRESS,
+                DDS_BOOLEAN_FALSE /* propagate */))
+              {
+                RMW_CONNEXT_LOG_ERROR_A_SET(
+                  "failed to assert property on participant: %s",
+                  "dds.transport.UDPv4.builtin.parent.allow_multicast_interfaces_list")
+                return RMW_RET_ERROR;
+              }
+          }
```
asorbini, line 256: I'm not familiar with this idiom, and I'm not sure how C++17 would change this block, but I definitely dislike this "fall through" block.

It can be simplified by using a boolean flag that is set separately in the block for the valid enums, and in the `default` one (which should be moved after)

```suggestion
          bool set_multicast_peers = false;
          switch (range) {
            case RMW_AUTOMATIC_DISCOVERY_RANGE_SUBNET:
              /* No action needed. This is the default discovery behavior for DDS */
              break;
            case RMW_AUTOMATIC_DISCOVERY_RANGE_DEFAULT:
            case RMW_AUTOMATIC_DISCOVERY_RANGE_LOCALHOST:
            case RMW_AUTOMATIC_DISCOVERY_RANGE_OFF:
              set_multicast_peers = false;
              break; 
            default:
              RMW_CONNEXT_LOG_WARNING_A(
                "Unknown value provided for automatic discovery range: %i",
                ctx->discovery_options->automatic_discovery_range);
              set_multicast_peers = false;
              break;
          }
          if (set_multicast_peers) {
            /* Note: We allow the LOCALHOST interface for the OFF range
               because if we leave this property completely blank then it
               has the opposite effect and allows all interfaces to be used.
               Allowing only LOCALHOST at least minimizes the unnecessary
               discovery traffic and prevents discovery with other host
               machines, while the domain_tag protects against same-host
               connections. */
            if (DDS_RETCODE_OK != DDS_PropertyQosPolicyHelper_assert_property(
                &dp_qos->property,
                "dds.transport.UDPv4.builtin.parent.allow_multicast_interfaces_list",
                RMW_CONNEXT_LOCALHOST_ONLY_ADDRESS,
                DDS_BOOLEAN_FALSE /* propagate */))
            {
              RMW_CONNEXT_LOG_ERROR_A_SET(
                "failed to assert property on participant: %s",
                "dds.transport.UDPv4.builtin.parent.allow_multicast_interfaces_list")
              return RMW_RET_ERROR;
            }
        }
```

asorbini (changes requested):
Thank you for this contribution, and apologies for the late reply. Things look good, except for a few minor changes (mostly "coding style" and missing error checking).

I'll review and suggest a solution to the problems with the failing cases of the matrix chart next.

rmw_connextdds_common/src/common/rmw_context.cpp
asorbini: After further review, this variable should be deprecated now that `ROS_STATIC_PEERS` is a thing, since they serve the same purpose.

We could add a note to the description in the README, but it can also be addressed by a separate PR.

asorbini:
> The gist of the current problem is:
> 
>     * when two endpoints are on different hosts
> 
>     * and one or both endpoints are in `LOCALHOST` mode
> 
>     * and neither endpoint is in `OFF` mode
> 
>     * and only one endpoint declares the other as a static peer
> 
> 
> -> then the required behavior is for them to discover each other and connect, but they currently do not. Every other scenario is working correctly.
> 
> From my reading of the Connext DDS documentation, I believe the current implementation is supposed to work since `accept_unknown_peers` is set to true, but perhaps there is a detail that I'm overlooking.

I've been reviewing the requirements, your implementation, and thinking about the problem. At first I thought I had spotted an issue, and started thinking about alternative implementations, but after going down that rabbit hole for a little bit, I realized I was wrong.

Your solution is the correct way to prevent the default multicast discovery from occurring while still allowing unicast communication to occur, and I have verified it with a simple hello world for Connext 6.0.1.

[discovery_options_tests.tar.gz](https://github.com/ros2/rmw_connextdds/files/11097245/discovery_options_tests.tar.gz) (I have only modified `ShapeType_publisher.c` and `ShapeType_subscriber.c` to customize `dds.transport.UDPv4.builtin.parent.allow_multicast_interfaces_list`).

I tried running the publisher on one host and the subscriber on a different one connected via LAN: by default, the processes don't discover each other, but if one declares the other's ip via `NDDS_DISCOVERY_PEERS`, then communication occurs as expected.

I will continue my testing by building the branch and observing the rmw's behavior.

rmw_connextdds_common/src/common/rmw_context.cpp
mxgrey: If @wjwwood was commenting on the unnecessary addition of the space, I think that was probably a fluke from my IDE's auto-formatter.

But I agree that the `RMW_CONNEXT_ENV_INITIAL_PEERS` env var could be deprecated now since it's redundant with the `ROS_STATIC_PEERS` env var.

rmw_connextdds_common/src/common/rmw_impl.cpp
mxgrey: It's indeed unrelated to the rest of the PR, but I was looking through this code for reference and noticed that it wasn't checking the correct pointer here, so I thought I'd slip in this fix.

asorbini:
@mxgrey @clalancette  I have identified the issue and implemented a solution [here](https://github.com/asorbini/rmw_connextdds/commit/f583d2ceed050ba69bb45cd94118d1c3cad9b65b) (the commit should apply directly on top of your branch).

Basically, the problem stems from the interaction between variable `NDDS_DISCOVERY_PEERS` and Connext's multicast receive address, explained in this [section of the Connext manual](https://community.rti.com/static/documentation/connext-dds/6.0.1/doc/manuals/connext_dds/html_files/RTI_ConnextDDS_CoreLibraries_UsersManual/index.htm#UsersManual/DISCOVERY_Qos.htm#9.5.2.4_Configuring_Multicast_Receive_Addresses%3FTocPath%3DPart%25202%253A%2520Core%2520Concepts%7C9.%2520Working%2520with%2520DDS%2520Domains%7C9.5%2520DomainParticipant%2520QosPolicies%7C9.5.2%2520DISCOVERY%2520QosPolicy%2520(DDS%2520Extension)%7C_____4). Quoting the comment I added in the code:

> Connext looks at variable NDDS_DISCOVERY_PEERS to determine whether it should add a multicast locator to the set of locators used for discovery. If this variable is empty, or if it contains at least one multicast address, a multicast locator is used for discovery (the default 239.255.0.1 group when the variable is empty, the first multicast locator found in the list when one or more multicast addresses are present).
> Because of this, we must make sure that NDDS_DISCOVERY_PEERS contains some value to prevent this default behavior, otherwise the participant will be announcing the default multicast group, which in turn will cause ROS_STATIC_PEERS not to work correctly

My proposed solution is to detect whether the variable is empty when the `LOCALHOST` discovery policy is used, and set it to some value (127.0.0.1 seems like a safe value) so that Connext will not automatically add the multicast address. The variable must be set before the DomainParticipantFactory is initialized, which forced me to move that initialization from `rmw_api_connextdds_ini` to `rmw_context_t::initialize_node`.

Let me know if this fixes the failing test cases in the matrix, and I'll keep an eye out for updates on this PR to have it merged.

wjwwood:
I cherry-picked your commit @asorbini, I'm still working on some of my own changes, then I'll ping you again. Thanks!

rmw_connextdds_common/src/common/rmw_context.cpp
```diff
@@ -435,6 +442,21 @@ rmw_context_impl_t::finalize()
 {
   rmw_ret_t rc_exit = RMW_RET_OK;
 
+  if (this->discovery_options && this->base) {
```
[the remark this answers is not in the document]
wjwwood: I fixed up this function according to your comments @asorbini in 5ae99e9.

However, I also removed the check about `this->base`, because it seemed to me that if `this->discovery_options` was not nullptr, then we should finalize it, regardless of base. @mxgrey can you comment on whether or not that's correct?

rmw_connextdds_common/src/common/rmw_context.cpp
```diff
@@ -435,6 +442,21 @@ rmw_context_impl_t::finalize()
...
+      this->base->options.allocator);
+    if (RMW_RET_OK != rc) {
+      RMW_CONNEXT_LOG_ERROR_A(
+        "failed to deallocate discovery options: %i",
+        rc);
+    }
+  }
+
+  if (this->domain_tag) {
+    DDS_String_free(this->domain_tag);
```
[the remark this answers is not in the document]
wjwwood: 5ae99e9

rmw_connextdds_common/src/common/rmw_context.cpp
wjwwood: That's right, I was commenting on the whitespace, sorry I should have been more specific.

As for the idea of deprecating the other env var, I'm fine with that too, but should we do that separately? Or would you prefer to do that in this pr @asorbini? I really think it's up to you.

rmw_connextdds_common/src/common/rmw_impl.cpp
```diff
@@ -205,6 +205,51 @@ rmw_connextdds_parse_string_list(
...
+  if (static_peers == nullptr) {
+    RMW_CONNEXT_LOG_ERROR_A_SET(
+      "received nullptr static_peers but a static_peer_count of %lu",
+      static_peer_count);
+    return RMW_RET_ERROR;
+  }
+
+  const auto initial_length = DDS_StringSeq_get_length(out);
+  const auto new_seq_length = initial_length + static_peer_count;
+  DDS_StringSeq_ensure_length(out, new_seq_length, new_seq_length);
```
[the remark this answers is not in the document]
wjwwood: I fixed this in 2d52d8f, @asorbini can you double check that?

rmw_connextdds_common/src/common/rmw_impl.cpp
wjwwood: I tried to address this in 9aef6d1, please check my work.

rmw_connextdds_common/src/ndds/dds_api_ndds.cpp
wjwwood: 43e6382

rmw_connextdds_common/src/ndds/dds_api_ndds.cpp
wjwwood: I didn't notice this comment until after I changed some things about the fallthrough stuff. But I attempted it in b4c9e9c.

wjwwood:
Ok, I think I'm finished addressing feedback, @asorbini can you have another look when you have time?

rmw_connextdds_common/src/common/rmw_impl.cpp
```diff
@@ -234,14 +237,14 @@ rmw_ret_t rmw_connextdds_extend_initial_peer_list(
          then free their memory. */
       DDS_String_free(*element_ref);
     }
-    const auto peer_char_len = strlen(peer);
-    *element_ref = DDS_String_alloc(peer_char_len);
+    const auto peer_char_len = strnlen(peer, RMW_DISCOVERY_OPTIONS_STATIC_PEERS_MAX_LENGTH);
+    *element_ref = DDS_String_alloc(peer_char_len + 1);  // +1 for nul character
```
asorbini, line 241: Minor, but `DDS_String_alloc` [already accounts for the nul terminator](https://community.rti.com/static/documentation/connext-dds/6.0.1/doc/api/connext_dds/api_c/group__DDSStringClass.html#gac243227428eb05799e9f1ab30aeefa85)

rmw_connextdds_common/src/ndds/dds_api_ndds.cpp
```diff
@@ -202,8 +203,17 @@ rmw_connextdds_initialize_participant_qos_impl(
 #endif /* RMW_CONNEXT_SHARE_DDS_ENTITIES_WITH_CPP */
   if (!ctx->domain_tag) {
     const auto pid = rcutils_get_pid();
-    ctx->domain_tag = DDS_String_alloc(18 + log10(pid));
-    sprintf(ctx->domain_tag, "ros_discovery_off_%d", pid);
+    static const char * format_string = "ros_discovery_off_%d";
+    int bytes_needed = rcutils_snprintf(nullptr, 0, format_string, pid);
+    ctx->domain_tag = DDS_String_alloc(bytes_needed + 1);
```
asorbini, line 208: Same here, the `+1` can be avoided

rmw_connextdds_common/src/ndds/dds_api_ndds.cpp
```diff
@@ -215,44 +225,42 @@ rmw_connextdds_initialize_participant_qos_impl(
...
           switch (range) {
             case RMW_AUTOMATIC_DISCOVERY_RANGE_SUBNET:
               /* No action needed. This is the default discovery behavior for DDS */
               break;
+            case RMW_AUTOMATIC_DISCOVERY_RANGE_DEFAULT:
+            case RMW_AUTOMATIC_DISCOVERY_RANGE_LOCALHOST:
+            case RMW_AUTOMATIC_DISCOVERY_RANGE_OFF:
+              set_multicast_peers = true;
+              break;
             default:
```
asorbini, line 238: Thinking a bit about this, why should we accept an invalid range type?

Is it to support future values? because if so, they wouldn't really be "supported", just "accepted" and treated like another value, but they most likely won't behave as expected.

I feel like it would be better to return an error so that a user knows the RMW will not behave as expected with the selected mode.

asorbini (approved):
Looks good, I think it can be merged with a green run of CI.

I only had a question about whether we should accept invalid range enums, which I'd like to address, and two other minor things with `DDS_String_alloc` (which could stay as is, but it would be great to clean up).

asorbini:
@wjwwood I realized that I was under the wrong impression that only `NDDS_DISCOVERY_PEERS` will affect the determination of whether a multicast receive address is announced or not, whereas just setting `initial_peers` will not achieve the same effect. 

After a bit more thinking, I believe they are equivalent, and I'd like to try reimplementing the feature with that strategy.

The reason for this is that it would be a cleaner implementation that could potentially allow two DomainParticipants to be created with different discovery options. This is not a problem at the moment, because the options are process-/ROS 2 context-/DDS factory-wide, but the change would likely make the implementation more future-proof in case we change things in the future.

It would also remove the new dependency from `rcutils_set_env`, which isn't much of a problem per se (since we already depend on `rcutils_get_env`), but "less dependencies"/"the same number of dependencies" is always a better scenario than "one more dependency" in my book.

The change would require moving back the participant factory initialization to `rmw_connextdds_init`, and checking if `ctx->initial_peers` is empty after asserting the static peers, and if so, filling it with "127.0.0.1" (when policy is LOCALHOST).

I can prepare the code and link a commit to cherry-pick, if that's ok with you. I should have it ready by the end of the day, and we can test/merge the PR tomorrow.

I'll start preparing it, while I wait for your input.

rmw_connextdds_common/src/common/rmw_context.cpp
asorbini: I think it's better to address that deprecation in a separate PR.

rmw_connextdds_common/src/common/rmw_context.cpp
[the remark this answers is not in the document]
asorbini: I noticed that and it was the right call imho.

wjwwood:
> After a bit more thinking, I believe they are equivalent, and I'd like to try reimplementing the feature with that strategy.

That is fine with me.

> It would also remove the new dependency from rcutils_set_env, which isn't much of a problem per se (since we already depend on rcutils_get_env), but "less dependencies"/"the same number of dependencies" is always a better scenario than "one more dependency" in my book.

Also fine with me.

> I can prepare the code and link a commit to cherry-pick, if that's ok with you. I should have it ready by the end of the day, and we can test/merge the PR tomorrow.

Sounds good to me, thank you for pushing on it so quickly! I'll wait to re-test the matrix of use cases until you have this.

asorbini:
@wjwwood After further testing, turns out I was right the first time around, and we must set `NDDS_DISCOVERY_PEERS` for the multicast receive address not to be used. Setting only `initial_peers` is not enough.

So I left the implementation as is, but I took the opportunity to consolidate all the changes into `rmw_context.cpp` so that they can be shared by `rmw_connextddsmicro` too (and we can think about deprecating `RMW_CONNEXT_INITIAL_PEERS` in favor of `ROS_STATIC_PEERS` in the near future).

I committed the changes [here](https://github.com/asorbini/rmw_connextdds/commit/6d4f3ea70abcf21ccd71e8dd065678885afa34be).

While at it, I also addressed those two comments I had about `DDS_String_alloc` already taking into account the `nul` terminator, and actually got rid of one of those two calls by remembering that `DDS_String_dup` exists (and so we don't need to use `DDS_String_alloc`+`memcpy`).

I also run the linter/uncrustify and fixed a couple of issues. Should be good for testing.

rmw_connextdds_common/src/common/rmw_context.cpp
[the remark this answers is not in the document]
mxgrey: Yeah, the check was there because if `base` was `nullptr` then there was no way to deallocate, but since the upstream API was changed to put the allocator inside of `rmw_discovery_options` itself, we don't depend on still having `base` anymore. LGTM.

asorbini:
I was just told that it might be possible to prevent the multicast receive behavior by resetting `DomainParticipantQos::discovery::multicast_receive_address` directly.

It might be a better idea, because of similar reasons I explained earlier (work at the participant level, instead of factory, don't mess with env vars., etc...).

I apologize for the back and forth, but I'm going to take a quick look into implementing this. While the current implementation works, it does have the problem that it prevents a unit test from initializing two contexts with two different policies, which might be a problem.

asorbini:
I applied and tested the change, and it does work as expected.

@wjwwood can you apply these two commits on top of your branch and run CI/test matrix?
- [b156654](https://github.com/asorbini/rmw_connextdds/commit/b1566548aa26a73ac739297ee4f64e737c500322) (the one I created yesterday, which refactors the logic to a separate function)
- [f5df932](https://github.com/asorbini/rmw_connextdds/commit/f5df9323e7eb76e7f59a91c6ac08b0567dcc7ef1) (the one from this morning, which resets `multicast_receive_addresses`)

The commits are the two most recent ones from this [branch](https://github.com/asorbini/rmw_connextdds/commits/discovery-peers-specification).

wjwwood:
I cherry-picked those and update it to not accept `NOT_SET` explicitly. I also added a doc note about this and we're bringing fastrtps and cyclone up to that too: https://github.com/ros2/rmw/pull/338/commits/2dbd6ef6d09d08dc0351ba0e88a99716db769a5a

rmw_connextdds_common/src/ndds/dds_api_ndds.cpp
wjwwood: We shouldn't, thanks for pointing this out. We are updating the other implementations too, and I cherry-picked your changes to make this the case in connext too.

rmw_connextdds_common/src/common/rmw_context.cpp
```diff
@@ -74,6 +76,202 @@ rmw_connextdds_initialize_participant_factory_qos(
...
+  switch (range) {
+    case RMW_AUTOMATIC_DISCOVERY_RANGE_SYSTEM_DEFAULT:
+    case RMW_AUTOMATIC_DISCOVERY_RANGE_SUBNET:
+      /* No action needed. This is the default discovery behavior for DDS */
+      break;
+    case RMW_AUTOMATIC_DISCOVERY_RANGE_NOT_SET:
+      RMW_CONNEXT_LOG_ERROR_SET(
+        "automatic_discovery_range unexpectedly RMW_AUTOMATIC_DISCOVERY_RANGE_NOT_SET");
+      return RMW_RET_ERROR;
+      break;
```
asorbini, line 142: The `break` here is unnecessary.

Also, we could skip this case in favor of `default which does the same thing, unless you prefer having an explicit message to detect `NOT_SET`.

rmw_connextdds_common/src/common/rmw_context.cpp
```diff
@@ -94,26 +292,39 @@ rmw_connextdds_initialize_participant_qos(
...
   }
 
   return RMW_RET_OK;
 }
 
 rmw_ret_t
 rmw_context_impl_t::initialize_node(
-  const rmw_discovery_options_t * const discovery_options)
+  const rmw_discovery_options_t * const discovery_options_in)
```
asorbini, line 321: Good call, thanks!

asorbini (changes requested):
Just a minor touch up, otherwise I think we're looking good

rmw_connextdds_common/src/common/rmw_context.cpp
wjwwood: Yeah, that's true.

I think an explicit message for NOT_SET is good to have.

rmw_connextdds_common/src/common/rmw_context.cpp
wjwwood: Yeah, not strictly necessary, but I found it helpful to make sure we didn't accidentally pass one where it should have been the other.

wjwwood:
I'm currently debugging an issue with the pr in it's current state. It's failing to create a node when using `RMW_IMPLEMENTATION=rmw_connextdds ROS_AUTOMATIC_DISCOVERY_RANGE=LOCALHOST ROS_STATIC_PEERS=127.0.0.1`.

asorbini:
> I'm currently debugging an issue with the pr in it's current state. It's failing to create a node when using `RMW_IMPLEMENTATION=rmw_connextdds ROS_AUTOMATIC_DISCOVERY_RANGE=LOCALHOST ROS_STATIC_PEERS=127.0.0.1`.

I'm not able to reproduce on my side. I'm rebuilding the tree after pulling all repositories to see if that makes a difference (looks like I was missing a few of the most recent commits).

wjwwood:
We figured out the issue and fixed it in [4bafd73](https://github.com/ros2/rmw_connextdds/pull/108/commits/4bafd738af4be7df834e93e246378593629cfebb). Thanks for your help @asorbini!

rmw_connextdds_common/src/common/rmw_context.cpp
wjwwood: [e3a5c8f](https://github.com/ros2/rmw_connextdds/pull/108/commits/e3a5c8fe74d2b4419fc89c345b0ef59d5e9513df)

asorbini (approved):
Good to go with green CI

rmw_connextdds_common/src/common/rmw_context.cpp
```diff
@@ -254,16 +254,36 @@ rmw_connextdds_initialize_discovery_options(
...
+      // This controls the number of participants that can be discovered on a single host,
+      // which is roughly equivalent to the number of ROS 2 processes.
+      // If it's too small then we won't connect to all participants.
+      // If it's too large then we will send a lot of announcement traffic.
+      // The default number here is picked arbitrarily.
+      const rmw_peer_address_t localhost_peers[2] = {
+        "32@builtin.udpv4://127.0.0.1",
+        "32@builtin.shmem://",
+      };
+      const auto rc2 = rmw_connextdds_extend_initial_peer_list(
```
asorbini, line 278: These peers should be added to the list only if it has not been already customized by the user in the environment variable and/or XML Qos, otherwise they might shadow custom values they configured

sloretz: Thats a good point. The current strategy for enabling users to control these settings is to set the range to `RMW_AUTOMATIC_DISCOVERY_RANGE_SYSTEM_DEFAULT`. It looks like maybe line 242 of this PR should be `else if (ctx->discovery_options->automatic_discovery_range != ..._SYSTEM_DEFAULT)` so that we don't try to extend the peer list in that case.

sloretz: Skipped modifying initial peers list when automatic discovery range is SYSTEM_DEFAULT in c501a76

PR #109, opened 2023-03-16
Event listeners distinct

cwecht:
Applies the fix suggested by @mauropasse here https://github.com/irobot-ros/events-executor/issues/16#issuecomment-1470262345


rmw_connextdds_common/include/rmw_connextdds/rmw_waitset_std.hpp
```diff
@@ -350,43 +350,44 @@ class RMW_Connext_StatusCondition : public RMW_Connext_Condition
...
 
 protected:
   DDS_StatusCondition * scond;
   std::mutex new_event_mutex_;
-  rmw_event_callback_t new_event_cb_{nullptr};
-  const void * user_data_{nullptr};
-  uint64_t unread_events_count_ = 0;
+  rmw_event_callback_t new_event_cb_[RMW_EVENT_INVALID] = {};
+  const void * user_data_[RMW_EVENT_INVALID] = {};
+  uint64_t unread_events_count_[RMW_EVENT_INVALID] = {0};
```
asorbini, line 390: The use of an enum value to define the size of an array is an unsual pattern which caught me by surprise.

It makes sense, but it's not commonly used, so I would add a comment to point it out.

Something like: "These arrays must be able to contain a value for each event kind. RMW_EVENT_INVALID is used because it is the last enumeration of rmw_event_type_t".

asorbini (approved):

PR #114, opened 2023-04-03
Only read unread samples count no callback has been set

cwecht:
Reading the count of unread samples should only be necessary in case there is now callback registered, as the count of unread samples is only read when a callback is set.

It turns out that couting the number of unread samples is a somewhat costly operation, so it should only be executed if necessary.


cwecht:
Ok, it seems, that this changes don't make any sense according to https://github.com/ros2/rmw_connextdds/pull/76/files#r832450207

PR #120, opened 2023-05-15
Add ros2_tracing tracepoints

cwecht:
Depends on https://github.com/ros2/ros2_tracing/pull/74; implements: https://github.com/ros2/ros2_tracing/issues/73

This PR introduces ros2_tracing tracepoints which are really helpful for examining system performance. The dependency to the ros2_tracing PR stems from the fact, that one of the original tracepoints had to be extended in order to get the timestamp during writing to DDS.

In order to do that, `DDS_DataWriter_write_untypedI` had to be replaced with  `DDS_DataWriter_write_w_timestamp_untypedI`. At first it seemed reasonable to use` in order to get the respective timestamp.

This seems to work, but `DDS_DomainParticipant_get_current_time` introduces a massive performance penality. The flamegraph below shows the result of profiling it using perf.
grafik

From https://community.rti.com/kb/how-clocks-are-used-connext-dds I'd conclude, that we need the "external clock" used by Connext here. Looking at the perf profile, it seems that there is a method/function `DDS_DomainParticipant_get_external_clockI` which would provide to us said clock, but it seems that this function is not part of Connexts public API right? Is there any way to get this clock or the timestamp which Connext would use if  `DDS_DataWriter_write_untypedI` is used without paying such a massive performance penalty?

### DIFF
--- a/rmw_connextdds_common/include/rmw_connextdds/dds_api.hpp
+++ b/rmw_connextdds_common/include/rmw_connextdds/dds_api.hpp
@@ -56,6 +56,11 @@ RMW_CONNEXTDDS_PUBLIC extern const char * const RMW_CONNEXTDDS_ID;
 extern const char * const RMW_CONNEXTDDS_SERIALIZATION_FORMAT;
 
 rmw_ret_t
+rmw_connextdds_get_current_time(
+  DDS_DomainParticipant * domain_participant,
+  struct DDS_Time_t * current_time);
+
+rmw_ret_t
 rmw_connextdds_set_log_verbosity(rmw_log_severity_t severity);
 
 rmw_ret_t
--- a/rmw_connextdds_common/include/rmw_connextdds/dds_api.hpp
+++ b/rmw_connextdds_common/include/rmw_connextdds/dds_api.hpp
@@ -49,7 +49,7 @@ enum RMW_Connext_MessageType
 struct RMW_Connext_WriteParams
 {
   DDS_Time_t timestamp{DDS_TIME_INVALID};
-  int64_t * sn_out{nullptr};
+  int64_t sequence_number{0};
 };
 
 RMW_CONNEXTDDS_PUBLIC extern const char * const RMW_CONNEXTDDS_ID;
--- a/rmw_connextdds_common/include/rmw_connextdds/dds_api.hpp
+++ b/rmw_connextdds_common/include/rmw_connextdds/dds_api.hpp
@@ -46,6 +46,12 @@ enum RMW_Connext_MessageType
   RMW_CONNEXT_MESSAGE_REPLY
 };
 
+struct RMW_Connext_WriteParams
+{
+  DDS_Time_t timestamp{DDS_TIME_INVALID};
+  int64_t * sn_out{nullptr};
+};
+
 RMW_CONNEXTDDS_PUBLIC extern const char * const RMW_CONNEXTDDS_ID;
 extern const char * const RMW_CONNEXTDDS_SERIALIZATION_FORMAT;
 
@@ -132,7 +138,7 @@ rmw_ret_t
 rmw_connextdds_write_message(
   RMW_Connext_Publisher * const pub,
   RMW_Connext_Message * const message,
-  int64_t * const sn_out);
+  RMW_Connext_WriteParams * const params = nullptr);
 
 rmw_ret_t
 rmw_connextdds_take_samples(
--- a/rmw_connextdds_common/src/common/rmw_impl.cpp
+++ b/rmw_connextdds_common/src/common/rmw_impl.cpp
@@ -1110,10 +1110,7 @@ rmw_connextdds_create_publisher(
     }
   }
 
-  TRACEPOINT(
-    rmw_publisher_init,
-    static_cast<const void *>(rmw_publisher),
-    rmw_pub_impl->gid()->data);
+  TRACETOOLS_TRACEPOINT(rmw_publisher_init, rmw_publisher, rmw_pub_impl->gid()->data);
 
   scope_exit_rmw_writer_impl_delete.cancel();
   scope_exit_rmw_writer_delete.cancel();
@@ -1931,10 +1928,7 @@ rmw_connextdds_create_subscriber(
     }
   }
 
-  TRACEPOINT(
-    rmw_subscription_init,
-    static_cast<const void *>(rmw_subscriber),
-    rmw_sub_impl->gid()->data);
+  TRACETOOLS_TRACEPOINT(rmw_subscription_init, rmw_subscriber, rmw_sub_impl->gid()->data);
 
 #if RMW_CONNEXT_DEBUG && RMW_CONNEXT_DDS_API == RMW_CONNEXT_DDS_API_PRO
   scope_exit_enable_participant_on_error.cancel();
@@ -2737,7 +2731,6 @@ RMW_Connext_Client::send_request(
 
 
   RMW_Connext_WriteParams write_params;
-  write_params.sn_out = sequence_id;
 
   if (DDS_RETCODE_OK !=
     DDS_DomainParticipant_get_current_time(
@@ -2749,6 +2742,8 @@ RMW_Connext_Client::send_request(
   }
 
   rmw_ret_t rc = this->request_pub->write(&rr_msg, false /* serialized */, &write_params);
+
+  *sequence_id = write_params.sequence_number;
 
   RMW_CONNEXT_LOG_DEBUG_A(
     "[%s] SENT REQUEST: "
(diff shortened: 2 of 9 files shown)


rmw_connextdds_common/src/common/rmw_impl.cpp
```diff
@@ -2973,6 +2994,14 @@ RMW_Connext_Service::send_response(
   rr_msg.gid.implementation_identifier = RMW_CONNEXTDDS_ID;
   rr_msg.payload = const_cast<void *>(ros_response);
 
+  DDS_Time_t timestamp;
```
asorbini, line 2997: Since we are using Connext's C API we should initialize the variable to a known state, e.g.

```suggestion
  DDS_Time_t timestamp = DDS_TIME_INVALID;
```

rmw_connextdds_common/src/common/rmw_impl.cpp
```diff
@@ -2973,6 +2994,14 @@ RMW_Connext_Service::send_response(
...
+  if (DDS_RETCODE_OK !=
+    DDS_DomainParticipant_get_current_time(this->reply_pub->dds_participant(), &timestamp))
+  {
+    RMW_CONNEXT_LOG_ERROR_SET("failed to write message to DDS")
```
asorbini, line 3001: I would change the message to be a bit more specific to this failure, e.g.

```suggestion
    RMW_CONNEXT_LOG_ERROR_SET("failed to get current time from DDS DomainParticipant")
```

asorbini (changes requested):
Hi @cwecht, thank you for your contribution! The changes look ok, I only had a few minor comments.

One thing to consider is that adding a single new argument to the `write()` functions is fine, but we might be venturing on a "slippery slope", and it might be better to refactor the functions to accept a single `struct` argument that contains all extra arguments. E.g.

  ```c
  // New structure for extra arguments. We could include the `serialized` flag too,
  // but that's already part of RMW_Connext_Message
  struct RMW_Connext_WriteParams {
    DDS_Time_t timestamp{DDS_TIME_INVALID};
    int64_t * sn_out{nullptr};
  };

  // Updated write() functions
  rmw_connextdds_write_message(
    RMW_Connext_Publisher * const pub,
    RMW_Connext_Message * const message,
    RMW_Connext_WriteParams * const params);

  class RMW_Connext_Publisher {
    write(
      const void * const ros_message,
      const bool serialized,
      RMW_Connext_WriteParams * consts params = nullptr);
  }
  ```

asorbini:
> From https://community.rti.com/kb/how-clocks-are-used-connext-dds I'd conclude, that we need the "external clock" used by Connext here. Looking at the perf profile, it seems that there is a method/function `DDS_DomainParticipant_get_external_clockI` which would provide to us said clock, but it seems that this function is not part of Connexts public API right? Is there any way to get this clock or the timestamp which Connext would use if `DDS_DataWriter_write_untypedI` is used without paying such a massive performance penalty?

I'm not very familiar with the clock configuration in Connext, and I'm looking into possible alternative solutions. I'll get back after I have a better idea of available options.

cwecht:
@asorbini Thank you for the review! I just applied the changes you requested and (forced) pushed them.

asorbini:
@cwecht are you sure you pushed the change? The code looks unchanged.

In general, I would prefer if you didn't force push but rather pushed an incremental commit so that the additional changes can be reviewed without having to re-scan through all of the code.

One reason to force push in this case would be to fix the commit message, which should include a DCO (e.g. "Signed-off-by: Your Name <your@email>"). I guess we are not strictly enforcing this rule and we are missing a CONTRIBUTING document in this repository (like [this one](https://github.com/ros2/demos/blob/rolling/CONTRIBUTING.md)), but the convention has been followed by all contributions, as far as I'm aware.

cwecht:
@asorbini now it's actually pushed including the signoff.

cwecht:
I've got some news regarding DDS_DomainParticipant_get_current_time.

I was able to test this PR with **connext micro** and using DDS_DomainParticipant_get_current_time in this case is just fine; there is basically no significant performance overhead.

For **Connext PRO** it is a different story:

1. The traces seem to indicate that the majority of the time in DDS_DomainParticipant_get_current_time is spend looking up the clock which is configured in the respective [QoS settings](https://community.rti.com/static/documentation/connext-dds/7.0.0/doc/manuals/connext_dds_professional/users_manual/users_manual/Clock_Selection.htm#domains_2763211417_1295556). The documentation seems to indicate that this has to be done since these QoS settings [may change at any time](https://community.rti.com/static/documentation/connext-dds/5.3.0/doc/manuals/connext_dds/html_files/RTI_ConnextDDS_CoreLibraries_UsersManual/Content/UsersManual/PROPERTY_Qos.htm#sending_2410472787_1561803)  which makes it reasonable to look up the clock which is supposed to be used every time.
2. However, if I understand the current implementation in rmw_connextdds correctly, the clock in use will always be the default clock, as PROPERY attribute of [DDS_DomainParticipantQos](http://community.rti.com/rti-doc/500/ndds.5.0.0/doc/html/api_c/structDDS__DomainParticipantQos.html) is never changed. From my understanding this can't even be changed via e.g. XML.
3. If the above is correct, we should be fine with using the system clock here directly as this should be always the clock which is used by DDS_DomainParticipant_get_current_time by default.

@asorbini could you double check this conclusion?

cwecht:
@asorbini I've added now `rmw_connextdds_get_current_time` which encapsulates the distinction between Pro and Micro regarding time retrieval. The performance difference between Pro's DDS_DomainParticipant_get_current_time and this one based on `RTIOsapiUtility_getTime` is quite impressive:  `RTIOsapiUtility_getTime` takes usually a few hunders of nanoseconds where `DDS_DomainParticipant_get_current_time` takes usually more like 6-20 microseconds. Do you have any objections regarding this approach?

cwecht:
@asorbini May I ask you kindly to have another look at this PR? Thank you very much!

asorbini:
Running full CI with [repos](https://gist.githubusercontent.com/asorbini/a23ad4d91288bb56c52e5c8f79330b60/raw/a7399429510dac58e1160d4acf1d9e205ae49a5a/rmw_connextdds_120.repos) for this PR, ros2/rmw_cyclonedds#454, ros2/rmw_fastrtps#694, ros2/ros2_tracing#74:

- Linux: [Build Status](https://ci.ros2.org/job/ci_linux/19984/)
- Linux aarch64: [Build Status](https://ci.ros2.org/job/ci_linux-aarch64/14452/)
- Windows: [Build Status](https://ci.ros2.org/job/ci_windows/20695/)

rmw_connextdds_common/include/rmw_connextdds/dds_api.hpp
```diff
@@ -46,9 +46,20 @@ enum RMW_Connext_MessageType
...
+  int64_t sequence_number{0};
+};
+
 RMW_CONNEXTDDS_PUBLIC extern const char * const RMW_CONNEXTDDS_ID;
 extern const char * const RMW_CONNEXTDDS_SERIALIZATION_FORMAT;
 
+rmw_ret_t
+rmw_connextdds_get_current_time(
+  DDS_DomainParticipant * domain_participant,
+  struct DDS_Time_t * current_time)
```
asorbini, line 61: @cwecht There's a missing `;` here which [causes the build to fail](https://ci.ros2.org/job/ci_linux/19984/console#console-section-10).

cwecht: Fixed.

asorbini:
Running CI again:

- Linux [Build Status](https://ci.ros2.org/job/ci_linux/19992/)
- Linux aarch64 [Build Status](https://ci.ros2.org/job/ci_linux-aarch64/14458/)
- Windows [Build Status](https://ci.ros2.org/job/ci_windows/20701/)

cwecht:
I'm note quite sure why the latest build failed. Looks more like a jenkins hiccup to me.

asorbini:
@cwecht agreed, seems like an unrelated failure. Running Linux again to check: [Build Status](https://ci.ros2.org/job/ci_linux/20004/).

The other two builds are yellow because of a formatting error unrelated to this PR.

clalancette:
> The other two builds are yellow because of a formatting error unrelated to this PR.

I will point out that the nightlies have no issues: https://ci.ros2.org/view/nightly/job/nightly_linux-aarch64_release/2582/ .  So it is worth investigating those before merging.

asorbini:
@clalancette thank you for the clarification. I totally agree that those issues should be resolved, but they are C++ linter errors and they stem from a different PR (ros2/rmw_cyclonedds#454), hence why I didn't think they'd affect this PR.

asorbini (approved):
CI passed (only failure comes from ros2/rmw_cyclonedds#454).

Changes can be merged once ros2/ros2_tracing#74 is in.

Issue #127, opened 2023-07-31
connext dds micro does not compile with rolling branch


cwecht:
**System Info**

- _OS_
  - Ubuntu 22.04
- *ROS version and installation type*
  - rolling/binary
- *RTI Connext DDS version and installation type*
  - Micro 3.0.3.28/rtipkg
- *RMW version or commit hash*
  - fff0668a7bc816a9075736a7c0ac08f124f394f9

**Bug Description**

rmw_connextdds_micro does not compile on rolling because of these errors:

![grafik](https://github.com/ros2/rmw_connextdds/assets/20225760/a815587e-ab83-419f-bba8-8fd0a609e813)

It seems that connext dds micro just doesn't support DDS_UserDataQosPolicy as well as the `multicast_receive_addresses` option of `DDS_DiscoveryQosPolicy`.

**Expected Behavior**

rmw_connextdds_micro should compile.

**How to Reproduce**

That's my setup:

1. setup a workspace with the following layout:
  - `rmw_connext_ws`
    - `rti_connext_dds-6.1.2` (containing the whole Pro package as well as the micro package in `rti_connext_dds_micro-3.0.3.28`)
    - `src/rmw_connextdds`
2. put `COLCON_IGNORE` in rmw_connext_ws/rti_connext_dds-6.1.2
3. run `docker run -it -v $PWD:/home -w /home osrf/ros:rolling-desktop in rmw_connext_ws`
4. run `source ./rti_connext_dds-6.1.2/resource/scripts/rtisetenv_x64Linux4gcc7.3.0.bash`
5. run `colcon build`

clalancette:
We don't officially support Connext Micro, so we don't usually test it.  For that matter, we don't even have licenses for Connext Micro.  If you have patches to make this compile, we'd be happy to review them.

fgallegosalido:
Hi @cwecht . Currently, we are not doing any further development on the Connext Micro RMW. As it is deprecated, I will close this issue.

PR #129, opened 2023-07-31
Pass parameters in the correct order to DDS_DataReader_read in rmw_co…

cwecht:
…nnextdds_count_unread_samples for micro.

The order of parameters is now consistent with the respective function call for Connext Pro.

### DIFF
--- a/rmw_connextdds_common/src/rtime/dds_api_rtime.cpp
+++ b/rmw_connextdds_common/src/rtime/dds_api_rtime.cpp
@@ -1220,8 +1220,8 @@ rmw_connextdds_count_unread_samples(
       &data_seq,
       &info_seq,
       DDS_LENGTH_UNLIMITED,
-      DDS_ANY_VIEW_STATE,
       DDS_NOT_READ_SAMPLE_STATE,
+      DDS_ANY_VIEW_STATE,
       DDS_ANY_INSTANCE_STATE);
     if (DDS_RETCODE_OK != rc && DDS_RETCODE_NO_DATA != rc) {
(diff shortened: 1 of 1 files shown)


clalancette (approved):
As far as I can tell, this has always been wrong; even on Connext Pro, the API seems to be:

```
DDS_DataReader_read(DDS_DataReader, DDS_UntypedSampleSeq, DDS_SampleInfoSeq, DDS_Long, DDS_SampleStateMask, DDS_ViewStateMask, DDS_InstanceStateMask)
```

So this looks good to me.  I'm going to run CI on it, but @asorbini I could really use a double-check here to make sure that we're doing the right thing with this change.

clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/19261/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/13804/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/19991/)

clalancette:
I'm going to go ahead and merge this since it seems right.  If there are any follow-up comments we can deal with them later.

cwecht:
This Fix really should be backported to humble.

clalancette:
> This Fix really should be backported to humble.

We can backport it to Humble (and Iron), but I'd rather wait a bit here to see if anything pops up in Rolling first.  While I don't *think* that will be the case, we did in fact change the bitmasks that are being passed.  So I want to make sure we didn't break anything.

cwecht:
@clalancette please not that due to #127 no one will be ablt to test these changes on the rolling branch since it won't even compile. This patch (and #130 as well BTW) will just work on humble as there is not build issue.

PR #131, opened 2023-09-06
Cleanup context implementation

clalancette:
I was running some tests under valgrind and I noticed that we had a memory leak here in `rmw_connextdds`.  As I took a closer look, it seemed to me that the initialization and finalization of contexts and nodes wasn't easy to follow, as well as leaking.

So this PR cleans that up and makes it a lot more straightforward.  It does a number of things:

1.  Hides a few methods in the implementation, rather than in the header file.
2.  Refactors the rmw_context_impl structure to have more `private` methods and members.
3.  Revamps how discovery options are handled, to fix the memory leak.

With this in place, I can no longer reproduce the memory leaks.


clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/19460/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/13983/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/20183/)

clalancette:
@asorbini your review here would be much appreciated.

rmw_connextdds_common/src/common/rmw_context.cpp
```diff
@@ -339,48 +361,9 @@ rmw_connextdds_initialize_participant_qos(
 }
 
 rmw_ret_t
-rmw_context_impl_t::initialize_node(
-  const rmw_discovery_options_t * const discovery_options_in)
+rmw_context_impl_s::initialize_node()
 {
-  if (this->node_count > 0) {
-    bool params_equal = false;
```
emersonknapp, line 346: This is the part that sticks out to me, these blocks seem to be just deleted, everything else in the diff seems to be just moved around/names changed.

I'm sure I just don't understand this code well enough, but could you explain why we no longer need this discovery options validation / initialization?

clalancette: Right, it's a good question, and is actually the heart of the fix here.

Prior to this PR, the code was actually doing two things with the discovery options that were wrong:
1.  It was allocating the discovery options when nodes were added.  But that really doesn't make any sense, since these discovery options are a per-context concept, not a per-node concept.  This is what led to the awkward code to check if discovery options are the same between nodes.  This is also what led to the memory leak, since in certain circumstances (adding one node, removing that node, and then adding another), we weren't properly freeing the associated memory.
2.  The discovery options were already stored in the `rmw_context_t` structure, in the `options.discovery_options` field, so there is no need to copy them at all here.  (I guess this isn't strictly "wrong", but is totally unnecessary)

To fix both of these, I changed it so that discovery options are always handled on a per-context basis.  That allows us to remove the code to check whether nodes in the same context have the same discovery options (since they do by definition), and also to get rid of the copy (since we just fetch it out of the `base_` member).

Does that make sense?

audrow (approved):
This is a big change, but seems to make sense with me.

Looks good to me with CI happy for all packages above 👍

emersonknapp (approved):

PR #132, opened 2023-09-25
Mitigate discovery race condition between clents and services

asorbini:
This PR introduces changes meant to reduce the impact of a race condition between clients and services.

The race condition is described in detail by [this comment](https://github.com/ros2/rmw_connextdds/blob/74d8d905fe3fe06c8080baac4f0c0874eb3ec791/rmw_connextdds_common/src/common/rmw_impl.cpp#L3119), but to summarize:

> It is possible for the service's request DataReader to receive a request from the client, handle it, and try to write the response before the service's reply DataWriter has actually matched the client's reply DataReader. If this is the case, the response will be lost.
> The data race and loss of reply message may also occur in the case of an "asymmetric discovery" between the client's reply DataReader and the service's reply DataWriter, where the latter has discovered and matched the former, but not viceversa.

The only way to fully resolve this race condition is described by the DDS-RPC 1.0 specification, section 7.6.2.2:

> [..] To ensure robustness, service discovery must ensure that no replies are lost due to race conditions in the discovery process of the underlying DDS entities [..]:
> - The client-side DataWriter (request DataWriter) shall not send out any request to a service DataReader unless the client-side DataReader (reply DataReader) has discovered and matched the corresponding service DataWriter (reply DataWriter).
> - The service-side DataReader (request DataReader) shall reject requests from a client DataWriter (request
DataWriter) as long as the service-side DataWriter (reply DataWriter) has not discovered and matched the
corresponding client-side DataReader (reply DataReader).

The proposed solution follows the [approach implemented by rmw_fastrtps_cpp](https://github.com/ros2/rmw_fastrtps/blob/b6c069cb1043041a2ac74ad7720b7a7ecfeb310f/rmw_fastrtps_shared_cpp/src/rmw_response.cpp#L129C1-L129C1):

- [The client (ab)uses the `related_sample_identity" metadata field of the request message to communicate its DataReader's GUID to the service](https://github.com/ros2/rmw_connextdds/blob/74d8d905fe3fe06c8080baac4f0c0874eb3ec791/rmw_connextdds_common/src/common/rmw_impl.cpp#L2803).
- [The service keeps track of discovered pairs of [DataReader, DataWriter] belonging to remote clients](https://github.com/ros2/rmw_connextdds/blob/74d8d905fe3fe06c8080baac4f0c0874eb3ec791/rmw_connextdds_common/include/rmw_connextdds/rmw_impl.hpp#L242).
- Before sending a reply back to the client, [the service waits for the client's DataReader to be matched locally, or throws a timeout error](https://github.com/ros2/rmw_connextdds/blob/74d8d905fe3fe06c8080baac4f0c0874eb3ec791/rmw_connextdds_common/src/common/rmw_impl.cpp#L3156) if that doesn't happen within the configured period ([read from DataWriterQos::reliability::max_blocking_time](https://github.com/ros2/rmw_connextdds/blob/74d8d905fe3fe06c8080baac4f0c0874eb3ec791/rmw_connextdds_common/src/common/rmw_impl.cpp#L1018)).

These changes, paired with the implementation of `rmw_service_server_is_available()`, provide a partial implementation of the solution prescribed by DDS-RPC, with the following limitations:
- `RMW_Connext_Client::::is_service_available()` only verifies that for every remote subscription matched by the local request DataWriter, there is a matched remote publication with the same "GUID prefix" (i.e. belonging to the same remote DomainParticipant) on the local reply DataReader. This test may lead to false positives if a remote DomainParticipant has created two or more "services" (i.e. pairs of request DR/reply DW) with the same (topic) name, e.g. in the case of a remote process with two ROS 2 Nodes that create a service with the same name. This is a very likely case for "1-to-many" services (where a client makes a request on many services at once).
- The client must use `rmw_service_server_is_available()` (e.g. by calling of `rclcpp::wait_for_service()`) before sending a request to make sure it'll (most likely) receive at least one reply from a service, but because of the false positives in `RMW_Connext_Client::is_service_available()`, it is still possible for the client to end up with "inconsistent" matches and no reply.
- A more robust solution requires each service to communicate their endpoints' GUIDs through discovery so that clients may correctly associate each pair of discovered endpoints and then make sure they are both matched before sending a request.
This information is currently encoded by the ROS Graph, by means of an endpoint's node and topic name, but it is not easily queried and correlated with the client's matches. It also relies on the assumption that a node can't create two services with the same name (not sure if that is the case...).


rmw_connextdds_common/src/common/rmw_impl.cpp
```diff
@@ -2662,9 +2751,11 @@ RMW_Connext_Client::take_response(
   if (taken_msg) {
     request_header->request_id.sequence_number = rr_msg.sn;
 
+    /* (asorbini) although messages are correlated using the reply DataReader's GUID,
+       we report the writer's GUID to upper layers. */
     memcpy(
       request_header->request_id.writer_guid,
-      rr_msg.gid.data,
+      this->request_pub->gid()->data,
```
asorbini, line 2758: I'm second guessing this change, and I'm thinking the value should be copied from `rr_msg.writer_gid.data`.

asorbini (commented):
Added some comments to help with reviewing.

I will now double-check the changes to `RMW_Connext_Client::take_response()`, then run CI.

rmw_connextdds_common/src/common/rmw_impl.cpp
asorbini: I [reviewed `rmw_take_response()`](https://github.com/ros2/rmw/blob/83445be486deae8c78d275e092eafb4bf380bd49/rmw/include/rmw/rmw.h#L2053), and the `writer_guid` is supposed to match the guid returned by `rmw_send_request()`, so the value returned here is the correct one.

asorbini:
[Full CI](https://ci.ros2.org/job/ci_launcher/12693/):

- Linux: [Build Status](https://ci.ros2.org/job/ci_linux/19555/)
- Linux aarch64: [Build Status](https://ci.ros2.org/job/ci_linux-aarch64/14068/) (interrupted because rmw_connextdds is disabled on this platform)
- Windows:  [Build Status](https://ci.ros2.org/job/ci_windows/20274/)

asorbini:
I'm trying to debug and resolve the Windows build failure in branch [asorbini/service-discovery-race-windbg](https://github.com/ros2/rmw_connextdds/tree/asorbini/service-discovery-race-windbg)

asorbini:
Resolved the Windows build issue (it was caused by an "out of order" `#include`).

[Running CI again](https://ci.ros2.org/job/ci_launcher/12697/):
- Linux: [Build Status](https://ci.ros2.org/job/ci_linux/19559/)
- Linux aarch64: [Build Status](https://ci.ros2.org/job/ci_linux-aarch64/14072/) (interrupted)
- Windows: [Build Status](https://ci.ros2.org/job/ci_windows/20283/)

asorbini:
CI passed with one unrelated test failure on Windows.

@clalancette is there someone that could take a look and approve the PR?

asorbini:
[Re-running CI](https://ci.ros2.org/job/ci_launcher/12750/) after resolving a deadlock caused by trying to query subscription on the DataWriter from within "matched_mutex":

- Linux [Build Status](https://ci.ros2.org/job/ci_linux/19620/)
- Linux aarch64 [Build Status](https://ci.ros2.org/job/ci_linux-aarch64/14128/) (interrupted)
- Windows [Build Status](https://ci.ros2.org/job/ci_windows/20342/)

bijoua29:
@clalancette Can this PR be reviewed as this has been open for a few months? We are running Connext rmw and need this fix in the release. Currently have to build this from source but would like to stop doing that.

bijoua29:
I understand the Jazzy rmw freeze is coming up soon? Can we get this PR reviewed and merged before then? It's been sitting here for 6 months.

fujitatomoya:
@bijoua29 i think this is really nice enhancement. as you noticed, RMW API freeze will be on Mon. April 8, 2024, https://docs.ros.org/en/rolling/Releases/Release-Jazzy-Jalisco.html#release-timeline

i will take a look sometime in this week, but this needs to be rebased before that.

CC: @asorbini

bijoua29:
> @bijoua29 i think this is really nice enhancement. as you noticed, RMW API freeze will be on Mon. April 8, 2024, https://docs.ros.org/en/rolling/Releases/Release-Jazzy-Jalisco.html#release-timeline
> 
> i will take a look sometime in this week, but this needs to be rebased before that.
> 
> CC: @asorbini

@fujitatomoya Thanks for reviewing this. 
@asorbini Can you rebase these changes?

clalancette:
I've rebased it onto the latest.

clalancette:
Here is full CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/20700/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/15015/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/21390/)

clalancette (approved):
I can't say I love that we can arbitrarily sleep during an `rmw_send_response`; that seems to me to be very jittery, and who knows if 100 milliseconds is long enough for this discovery to happen.  But it does improve the situation here, and also matches what `rmw_fastrtps` does, so I'm OK with it for now.

fujitatomoya:
@clalancette i was going to have a review, thanks!

clalancette:
OK, yeah.  It looks like this PR is the cause of the regressions.  I'm going to open a PR to revert it.

PR #133, opened 2023-10-06
Fix a couple of warnings pointed out by clang.

clalancette:
1.  Add braces around initialization of array elements.
2.  Remove an unused structure member.


asorbini (approved):
Looks good to me. Thank you for catching these issues!

PR #137, opened 2023-11-21
prevent setting CONNEXTDDS_ARCH when a more proper one exists

WanruXX:
The original line 519-523 may result in an improper architect is selected while a proper one can exist.

For example, if you have 
**CMAKE_HOST_SYSTEM_NAME=Linux
CMAKE_HOST_SYSTEM_PROCESSOR=x86_64
architectures_installed=[armv8Linux4.4gcc5.4.0;x64Linux4gcc7.3.0]**

the original cmake will set CONNEXTDDS_ARCH as **armv8Linux4.4gcc5.4.0** and break the foreach loop in the first entry without looking at the latter possible architects.

After the change,  **armv8Linux4.4gcc5.4.0** will be put into a _architecture_candidates_ list and continue searching, if **x64Linux4gcc7.3.0** can be found then it will select this one, if not it will select **armv8Linux4.4gcc5.4.0** as the best choice.

### DIFF
--- a/rti_connext_dds_cmake_module/cmake/rti_build_helper.cmake
+++ b/rti_connext_dds_cmake_module/cmake/rti_build_helper.cmake
@@ -476,6 +476,7 @@ function(rti_guess_connextdds_arch)
     message(STATUS
       "Pick first from ${CONNEXTDDS_DIR}/lib/[${architectures_installed}]")
 
+    set(architecture_candidates "")
     foreach(architecture_name ${architectures_installed})
       # Because the lib folder contains both target libraries and
       # Java JAR files, here we exclude the "java" in our algorithm
@@ -516,11 +517,7 @@ function(rti_guess_connextdds_arch)
           endif()
         endif()
         if(NOT CONNEXTDDS_ARCH)
-          message(STATUS
-            "unsupported CMAKE_HOST_SYSTEM_NAME (${CMAKE_HOST_SYSTEM_NAME}) "
-            "or CMAKE_HOST_SYSTEM_PROCESSOR (${CMAKE_HOST_SYSTEM_PROCESSOR}). "
-            "Using architecture ${architecture_name} anyway.")
-          set(CONNEXTDDS_ARCH "${architecture_name}")
+          list(APPEND architecture_candidates ${architecture_name})
         endif()
       else()
         message(STATUS "ignored foreign architecture: ${architecture_name}")
@@ -529,18 +526,25 @@ function(rti_guess_connextdds_arch)
       if(CONNEXTDDS_ARCH)
         break()
       endif()
+
     endforeach()
-  endif()
 
-  if(NOT CONNEXTDDS_ARCH)
-    message(WARNING
-      "CONNEXTDDS_ARCH not specified. Please set "
-      "-DCONNEXTDDS_ARCH= to specify your RTI Connext DDS "
-      " architecture")
-  else()
-      message(STATUS "Selected CONNEXTDDS_ARCH: ${CONNEXTDDS_ARCH}")
-  endif()
+    if(NOT CONNEXTDDS_ARCH)
+      list(GET architecture_candidates 0 CONNEXTDDS_ARCH)
+      if(CONNEXTDDS_ARCH)
+        message(STATUS
+          "unsupported CMAKE_HOST_SYSTEM_NAME (${CMAKE_HOST_SYSTEM_NAME}) "
+          "or CMAKE_HOST_SYSTEM_PROCESSOR (${CMAKE_HOST_SYSTEM_PROCESSOR}). "
+          "Using architecture ${CONNEXTDDS_ARCH} anyway.")
+      else()
+        message(WARNING
+          "CONNEXTDDS_ARCH not specified. Please set "
+          "-DCONNEXTDDS_ARCH= to specify your RTI Connext DDS "
+          " architecture")
+      endif()
+    endif()
 
+  endif()
 
   set(CONNEXTDDS_ARCH "${CONNEXTDDS_ARCH}" PARENT_SCOPE)
 endfunction()


rti_connext_dds_cmake_module/cmake/rti_build_helper.cmake
```diff
@@ -516,11 +517,7 @@ function(rti_guess_connextdds_arch)
           endif()
         endif()
         if(NOT CONNEXTDDS_ARCH)
-          message(STATUS
-            "unsupported CMAKE_HOST_SYSTEM_NAME (${CMAKE_HOST_SYSTEM_NAME}) "
-            "or CMAKE_HOST_SYSTEM_PROCESSOR (${CMAKE_HOST_SYSTEM_PROCESSOR}). "
-            "Using architecture ${architecture_name} anyway.")
-          set(CONNEXTDDS_ARCH "${architecture_name}")
+          list(APPEND architecture_candidates ${architecture_name})
```
fujitatomoya, line 520: i think they are not candidates any more, they are `unsupported_architecture` (the only difference is that does not have `java` in the list.) runtime library list for the host system?

rti_connext_dds_cmake_module/cmake/rti_build_helper.cmake
```diff
@@ -529,18 +526,25 @@ function(rti_guess_connextdds_arch)
...
+          "unsupported CMAKE_HOST_SYSTEM_NAME (${CMAKE_HOST_SYSTEM_NAME}) "
+          "or CMAKE_HOST_SYSTEM_PROCESSOR (${CMAKE_HOST_SYSTEM_PROCESSOR}). "
+          "Using architecture ${CONNEXTDDS_ARCH} anyway.")
+      else()
+        message(WARNING
+          "CONNEXTDDS_ARCH not specified. Please set "
+          "-DCONNEXTDDS_ARCH= to specify your RTI Connext DDS "
+          " architecture")
+      endif()
+    endif()
```
fujitatomoya, line 545: this message is missing after the fix?

```suggestion
    else()
      message(STATUS "Selected CONNEXTDDS_ARCH: ${CONNEXTDDS_ARCH}")
    endif()
```

rti_connext_dds_cmake_module/cmake/rti_build_helper.cmake
```diff
@@ -476,6 +476,7 @@ function(rti_guess_connextdds_arch)
     message(STATUS
       "Pick first from ${CONNEXTDDS_DIR}/lib/[${architectures_installed}]")
```
fujitatomoya, line 477:
```suggestion
    message(STATUS
      "Try to search from ${CONNEXTDDS_DIR}/lib/[${architectures_installed}]")
```

fujitatomoya (commented):
overall, this is enhancement. besides the comments in the implementation, i have a question.

what could be the situation that we might have, `architectures_installed=[armv8Linux4.4gcc5.4.0;x64Linux4gcc7.3.0]`?

rti_connext_dds_cmake_module/cmake/rti_build_helper.cmake
WanruXX: yeah i think they should be just unsopported too, i did this since the original code would set this unsopported architecture anyway, so i called them candidates...

rti_connext_dds_cmake_module/cmake/rti_build_helper.cmake
WanruXX: ohh i forgot it, thanks!

WanruXX:
theoratically we shouldn't have that situation... but unfortunately i don't know how the platform team in our department installed our rti that we have multiple libraries installed in our development environment for 300 developers😅
image

fujitatomoya (approved):
the actual use case is unclear for me, could be cross-build environment or something. (I see QNX! nice!) originally this cmake helper tries to find the appropriate library as much as possible, so i guess this can be better enhancement. lgtm

rti_connext_dds_cmake_module/cmake/rti_build_helper.cmake
```diff
@@ -529,18 +526,27 @@ function(rti_guess_connextdds_arch)
...
+          "or CMAKE_HOST_SYSTEM_PROCESSOR (${CMAKE_HOST_SYSTEM_PROCESSOR}). "
+          "Using architecture ${CONNEXTDDS_ARCH} anyway.")
+      else()
+        message(WARNING
+          "CONNEXTDDS_ARCH not specified. Please set "
+          "-DCONNEXTDDS_ARCH= to specify your RTI Connext DDS "
+          " architecture")
+      endif()
+    else()
       message(STATUS "Selected CONNEXTDDS_ARCH: ${CONNEXTDDS_ARCH}")
```
iuhilnehc-ynos, line 546: I am afraid that this status message could not be output if the [guessed_architecture](https://github.com/ros2/rmw_connextdds/blob/609f3ef699757b2305ea129b236b318415b41689/rti_connext_dds_cmake_module/cmake/rti_build_helper.cmake#L463) is matched, but the original logic did.

rti_connext_dds_cmake_module/cmake/rti_build_helper.cmake
```diff
@@ -529,18 +526,27 @@ function(rti_guess_connextdds_arch)
...
-  endif()
 
-  if(NOT CONNEXTDDS_ARCH)
-    message(WARNING
-      "CONNEXTDDS_ARCH not specified. Please set "
-      "-DCONNEXTDDS_ARCH= to specify your RTI Connext DDS "
-      " architecture")
-  else()
+    if(NOT CONNEXTDDS_ARCH)
+      list(GET architecture_candidates 0 CONNEXTDDS_ARCH)
```
iuhilnehc-ynos, line 533: what if the `architectures_installed` is empty, is it safe to get the item from `architecture_candidates`?

rti_connext_dds_cmake_module/cmake/rti_build_helper.cmake
```diff
@@ -462,7 +462,7 @@ function(rti_guess_connextdds_arch)
   if(EXISTS "${CONNEXTDDS_DIR}/lib/${guessed_architecture}")
     set(CONNEXTDDS_ARCH "${guessed_architecture}")
     message(STATUS
-      "Guessed ${CONNEXTDDS_DIR}/lib/${guessed_architecture} exists")
+      "Guessed ${CONNEXTDDS_DIR}/lib/${guessed_architecture} exists and is selected")
```
WanruXX, line 465: i changed the message here dince it will skip the message in line 546

rti_connext_dds_cmake_module/cmake/rti_build_helper.cmake
WanruXX: indeed, i modified the message in the case the guessed_architecture exists, please refer to below

rti_connext_dds_cmake_module/cmake/rti_build_helper.cmake
WanruXX: in this case the CONNEXTDDS_ARCH will be not set so will print out the message in line 540, which means the user didn't install any non-foreign architectures that would fit their operation system. The build will fail in this case and the original logical both.

iuhilnehc-ynos: I don't think the following error message is expected.

```
CMake Error at CMakeLists.txt:13 (list):
  list GET given empty list
```

WanruXX: sorry for not examining carefully, i made the change for checking it

PR #140, opened 2024-02-06
Add DDS_TIME_INVALID and DDS_Time_is_invalid for connext micro

cwecht:
For whatever reasons ConnextDDS micro exports DDS_TIME_INVALID but it doesn't define it. Therefore this results in a runtime error - because during runtime as the shared library is loaded, the symbol to DDS_TIME_INVALID can not be resolved.

Also DDS_Time_is_invalid is not defined. So both need to be added in order to make rmw_connextmicro work.

### DIFF
--- a/rmw_connextdds_common/include/rmw_connextdds/dds_api.hpp
+++ b/rmw_connextdds_common/include/rmw_connextdds/dds_api.hpp
@@ -48,8 +48,8 @@ enum RMW_Connext_MessageType
 
 struct RMW_Connext_WriteParams
 {
-  DDS_Time_t timestamp{DDS_TIME_INVALID};
-  int64_t sequence_number{0};
+  DDS_Time_t timestamp = DDS_TIME_INVALID;
+  int64_t sequence_number = 0L;
 };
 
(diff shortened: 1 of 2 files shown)


cwecht:
This PR is basically a direct follow up to #120. At the time I only checked whether rmw_connextmicro build successfully. Later I discuvered this issue with Connext Micro exporting symbols which it doesn't define.

@asorbini it would be great if you could have a look at this.

fujitatomoya (commented):
@cwecht it really looks like the problem is with ConnextDDS micro where it should address your issue here? Having the work-around to bypass the problem would be unexpected behavior in the future, since all these definition need to be in the DDS implementation?

@asorbini it would be appreciated if you can chime in here.

cwecht:
@asorbini I found neither for ConnextDDS Micro 3.0.3.28 nor for 2.4.13.4 any definition for DDS_TIME_INVALID.

fgallegosalido:
Hi @cwecht . Currently, we are not doing any further development on the Connext Micro RMW. As it is deprecated, I will close this pull request.

PR #141, opened 2024-02-29
Support Fast CDR v2

MiguelCompany:
This makes the necessary changes to allow building with Fast CDR v2

Part of ros2/ros2#1530


clalancette (approved):
This looks OK to me with green CI.

PR #144, opened 2024-03-12
Fixed naming convention of service response topics in documentation

cwecht:
The documentation of the topic naming convention for service responses has been wrong. In fact the suffix is not "Response" but "Reply":

- https://github.com/ros2/rmw_connextdds/blob/eebc271fa1420d1b9135a56f4507fe134460f39d/rmw_connextdds_common/src/common/rmw_info.cpp#L328
- https://github.com/ros2/rmw_cyclonedds/blob/d8973e818db1546975478dc59699f78417b9051b/rmw_cyclonedds_cpp/src/rmw_node.cpp#L4948


clalancette (approved):
Yep, good call.

Given that this is just a change to the README, I'm going to go ahead and merge it without CI.  Thanks for the fix.

PR #146, opened 2024-04-06
Revert "Mitigate discovery race condition between clents and services"

clalancette:
Reverts ros2/rmw_connextdds#132

This seems to be causing regressions on Windows; see https://github.com/ros2/rmw_connextdds/pull/132#issuecomment-2040348605 .

CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/20771/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/15070/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/21451/)
* Windows Debug [Build Status](http://ci.ros2.org/job/ci_windows/21452/)

PR #147, opened 2024-04-06
Revert "Revert "Mitigate discovery race condition between clents and services""

clalancette:
Reverts ros2/rmw_connextdds#146

This essentially reapplies https://github.com/ros2/rmw_connextdds/pull/132 , but it is a draft for now because there is clearly some kind of UB happening here that shows up on Windows Debug.  Needs to be debugged and fixed before we can put this back in.


trubio-rti:
Thank you for the draft PR. I've opened https://github.com/ros2/rmw_connextdds/pull/153 to follow up with the work, so I'm closing this.

PR #164, opened 2024-12-02
Remove the mirror-rolling-to-master workflow.

clalancette:
We no longer need to keep that compatibility, so just drop it.


christophebedard (approved):

ahcorde:
https://github.com/Mergifyio backport jazzy humble

PR #189, opened 2025-05-01
Removed deprecated connextddsmicro

ahcorde:
Open this PR as a draft because I have some questions:

 - `rmw_connextdds_common/CMakeLists.txt` contains many references to connextddsmicro. Should we remove them ?
 - `rti_connext_dds_cmake_module/env_hook/rti_connext_dds_micro.sh.in` should we remove this file?


fgallegosalido:
Hi @ahcorde, the idea is to do this in multiple steps:

1. Mark the `rmw_connextddsmicro` as deprecated (Done).
2. Decouple the `rmw_connextdds` package from the common one and also deprecate the `rmw_connextdds_common` package.
3. Remove both `rmw_connextdds_common` and `rmw_connextddsmicro` packages.

We don't want to rush this yet. We want to keep the package `rmw_conenxtddsmicro` available until we feel it is okay to remove it, but also allowing us to continue developing the `rmw_connextdds` package without all the Micro overhead.

PR #216, opened 2025-12-17
Remove default from switch with enum, so that compiler warns.

fujitatomoya:
## Description

Now the compiler will warn us if new enum values are added to it but aren't handled in this function.

Fixes # (issue)

### DIFF
--- a/rmw_connextdds/src/rmw_api_impl_ndds.cpp
+++ b/rmw_connextdds/src/rmw_api_impl_ndds.cpp
@@ -1033,11 +1033,13 @@ rmw_feature_supported(rmw_feature_t feature)
 {
   switch (feature) {
     case RMW_FEATURE_MESSAGE_INFO_RECEPTION_SEQUENCE_NUMBER:
+      [[fallthrough]];
     case RMW_FEATURE_MESSAGE_INFO_PUBLICATION_SEQUENCE_NUMBER:
       {
         return true;
       }
     case RMW_MIDDLEWARE_SUPPORTS_TYPE_DISCOVERY:
+      [[fallthrough]];
     case RMW_MIDDLEWARE_CAN_TAKE_DYNAMIC_MESSAGE:
       {
         return false;
--- a/rmw_connextdds/src/rmw_api_impl_ndds.cpp
+++ b/rmw_connextdds/src/rmw_api_impl_ndds.cpp
@@ -1037,11 +1037,13 @@ rmw_feature_supported(rmw_feature_t feature)
       {
         return true;
       }
-    default:
+    case RMW_MIDDLEWARE_SUPPORTS_TYPE_DISCOVERY:
+    case RMW_MIDDLEWARE_CAN_TAKE_DYNAMIC_MESSAGE:
       {
         return false;
       }
   }
+  return false;
 }
 
 /******************************************************************************
--- a/rmw_connextdds_common/src/common/rmw_impl.cpp
+++ b/rmw_connextdds_common/src/common/rmw_impl.cpp
@@ -352,8 +352,7 @@ rmw_connextdds_get_readerwriter_qos(
         history->kind = DDS_KEEP_ALL_HISTORY_QOS;
         break;
       }
-    // case RMW_QOS_POLICY_HISTORY_UNKNOWN:
-    default:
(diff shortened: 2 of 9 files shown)


fujitatomoya (commented):
@fgallegosalido minor enhancement to take advantage of compiler warning, can you take a look?

ahcorde:
Pulls: ros2/rmw_connextdds#216
Gist: https://gist.githubusercontent.com/ahcorde/f2ae7ba61018c7000ec7dadc8e8631ac/raw/593cd5eef89ff520f617a395c0d97ac08c911821/ros2.repos
BUILD args:  --packages-above-and-dependencies rmw_connextdds_common
TEST args:  --packages-above rmw_connextdds_common
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/17817
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/26676/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/20249/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/6486/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/26265/)

rmw_connextdds/src/rmw_api_impl_ndds.cpp
```diff
@@ -1037,11 +1037,13 @@ rmw_feature_supported(rmw_feature_t feature)
       {
         return true;
       }
-    default:
+    case RMW_MIDDLEWARE_SUPPORTS_TYPE_DISCOVERY:
```
fgallegosalido, line 1041: Apply this change in other places.
```suggestion
    case RMW_MIDDLEWARE_SUPPORTS_TYPE_DISCOVERY: [[fallthrough]]
```

rmw_connextdds_common/src/common/rmw_impl.cpp
```diff
@@ -442,8 +441,8 @@ rmw_connextdds_get_readerwriter_qos(
         liveliness->kind = DDS_MANUAL_BY_TOPIC_LIVELINESS_QOS;
         break;
       }
-    // case RMW_QOS_POLICY_LIVELINESS_UNKNOWN:
-    default:
```
fgallegosalido, line 446: Is there any possibility that a user can pass an invalid value for these enums (like some random value that doesn't match any name in the enum)? This question goes for other switchs without a default clause.

fgallegosalido (changes requested):

rmw_connextdds/src/rmw_api_impl_ndds.cpp
fujitatomoya: good practice to let the compiler know we are NOT missing `break` in this case. thanks 👍

rmw_connextdds_common/src/common/rmw_impl.cpp
fujitatomoya: enum data type can be checked during compile time, that means if user is using different data type of enum which is expected to be used in switch statement, that should be detected by the compiler during build.

reference: https://github.com/ros2/rmw_cyclonedds/pull/518#pullrequestreview-2374265975

fujitatomoya: if you prefer to keep the default, that is also fine for me. but we could miss the new cases in the future, that could lead the undefined behavior as well...

fujitatomoya:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/26697/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/20273/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/6496/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/26278/)

rmw_connextdds_common/src/common/rmw_impl.cpp
fgallegosalido: I don't like the default clause neither if you are being exhaustive. I'm okay with this

ahcorde (approved):

fujitatomoya:
@ahcorde @fgallegosalido thanks for the review and feedback!

